Issue #173, opened 2013-04-05
Add the ability to pass arguments to the script itself


glennblock:
I want to write a script that uses powerargs to parse cmdline params!

PR #228, opened 2013-05-07
Implement `Args` to pass arguments to scripts.

sblom:
Fix for #173

You call `scriptcs.exe` with arguments for it before a `--` and arguments for the script after a `--`.

For example:

```
scriptcs -restore server.csx -- -port 8080
```

Use it like:

``` csharp
Console.WriteLine("Args: '{0}'", string.Join("', '", Args));
```

(This would print `Args: '-port', '8080'`.)

### DIFF
--- a/src/ScriptCs/Command/ExecuteReplCommand.cs
+++ b/src/ScriptCs/Command/ExecuteReplCommand.cs
@@ -20,6 +20,8 @@ namespace ScriptCs.Command
         private readonly ILog _logger;
         private readonly IConsole _console;
 
+        public string[] ScriptArgs { get; private set; }
+
         public ExecuteReplCommand(
             IFileSystem fileSystem,
             IScriptPackResolver scriptPackResolver,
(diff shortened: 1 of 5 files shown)


hassanselim0:
Have this feature been removed or re-implemented differently?
(sorry for commenting on an old PR)

khellang:
It's now called `Env.ScriptArgs`. See [`IScriptHost`](https://github.com/scriptcs/scriptcs/blob/281b55ef5d24844dfefb91daeb4111efda264a52/src/ScriptCs.Contracts/IScriptHost.cs) and [`IScriptEnvironment`](https://github.com/scriptcs/scriptcs/blob/281b55ef5d24844dfefb91daeb4111efda264a52/src/ScriptCs.Contracts/IScriptEnvironment.cs) :smile:

adamralph:
FYI that change was made in version 0.9 - https://github.com/scriptcs/scriptcs/issues/279

PR #590, opened 2014-03-02
Word completion in REPL

estoy:
This PR adds completion of file paths and assembly names in the REPL.

New functionality:

#load gets file path completion. Just press tab any number of times and you're taken through the current possible completions of your path. Shift tab moves back through the list. Escape cancels the completion and restores the original input line.

#r gets assembly name completion. Press tab to loop through all possisble completions that are found in the GAC and in your current directory.

:cwd prints the current working directory

:cd <path> changes directory. It of course also offers file path completion.

Known issues:
Some of the assemblies found in the GAC directories cannot be loaded. I'm no GAC expert so any advice is appreciated.

Further improvements:
"Intellisense"-style completion of expressions. This is what I really wanted to achieve. I have all the infrustructure in place and Roslyn provides the semantic analysis that I need but I cannot find a way to access this so that it is done within the current session. Any ideas on how to achieve this?

Implementation structure:
The main class of the solution is the InputLine which takes over the read line responsibility from the console. It uses the following supporting classes:
ReplBuffer: Keeps track of the contents of the current line and updates the console to reflect this.
ReplHistory: Provides input line history, which is accessed through the up/down arrows.
LineAnalyzer: Determines what kind of input is in progress, e.g. file path or assembly name. Can be expanded simply by adding new regular expressions.
CompletionHandler: Issues searches for possible completions of file path, etc, and keeps track of which to show at the moment.
FilePathFinder: Finds possible file paths and assembly paths.

Effects on original code:
The main change is to the IConsole / ScriptConsole which have been augmented to expose several methods and properties of System.Console.

So, I hope this will prove to be useful. This is my first ever PR so please bear with me if I've done something unwise.


adamralph:
@estoy thanks very much for the work on this.

Due to delays on both sides, this PR has been open for a very long time now and the codebase has diverged significantly, e.g. REPL commands such cd: are already in master. I think the best thing now is if we close this, and spin off the features which it was adding as independent issues so we can address them in bite size chunks. I've raised #929, #930 and #931 for this purpose.

I've also pushed your spike branch into the upstream repo so it can be used as a reference when implementing those issues - https://github.com/scriptcs/scriptcs/tree/wordcompletion-spike

Again, due to the sheer amount of divergence, I don't think any promises can be made about any of your original commits being used, but I'm sure the ideas and approaches your spike demonstrates are going to be very useful indeed. And of course, if you are still motivated, you are more than welcome to join in in getting the 3 new issues closed.

src/ScriptCs.Contracts/IConsole.cs
```diff
@@ -19,5 +23,11 @@ public interface IConsole
         void ResetColor();
 
         ConsoleColor ForegroundColor { get; set; }
+
+        int HorizontalPosition { get; set; }
+
+        int VerticalPosition { get; set; }
+
+        int Width { get; }
```
BlackFrog1: I would like this property added to the IConsole interface also.  It's helpful when formatting output.

PR #746, opened 2014-06-02
Spike: Emit a host for Mono with static accessors for ScriptHost

stirno:
This is really a first-stab POC on getting custom `ScriptHost` impls working in Mono. Goal is to crawl the passed in `ScriptHost` and create a class with static proxy methods/accessors that can be used by the Mono engine.

Not perfect but it does work for PVC's `ScriptHost` in mono. Code needs some cleanup for sure, possibly doing something horribly wrong. Thoughts? :+1:


glennblock:
@stirno I am closing this as it has not had any progress in 2 years.

Issue #665, opened 2014-05-06
Mono REPL multiline


filipw:
Add multiline support for mono REPL
Currently the REPL for mono is single line only

I have a working spike but it needs some work. Unfortunately looks like the approach from Roslyn with catching compiler error might not be that simple to rebuild there.

---

**Edit by @adamralph**

This feature supports everything apart from loose functions. That has been spun off into https://github.com/scriptcs/scriptcs/issues/906

matthewblott:
Just started using this on OSX and noticed this issue - it looks pretty old so I'm guessing this isn't being worked on?

adamralph:
@matthewblott any Mono related work has indeed slipped down the order of importance somewhat. This is largely due to the fact that Microsoft seem committed to x-plat with their latest bits, so the hope is that, in the not too distant future, we will have a build of Roslyn that its fully x-plat and has the scripting API's reintroduced, which would allow us to kill the scriptcs Mono engine. Of course, we don't know when or indeed if that will definitely happen, so I guess Mono related scriptcs work is a little in limbo right now.

filipw:
mono REPL multiline was fixed long time ago by @ztone in this PR #771 
however we have done a terrible job managing those and as a result things have kinda slipped away.

@ztone we discussed extracting mono multiline REPL into a single PR a while ago - can this be PR-ed separately or is it relying on the changes introduced in #771? If that's the case, let's rebase and merge it ASAP. Very sorry it was left hanging for so long

ztone:
@filipw @adamralph @matthewblott  Tbh I would not want to merge #771 unless we really need to.  Only thing it fixes is the block parsing issue no one has actually complained about (yet!). Regarding issue #713, which #771 fixes as well, will be fixed when we get a new Mono.CSharp release because I found the issue in the mono compiler a few months ago.

So, getting back to this issue.  Adding multi-line support for everything but loose function is really easy (only 3 lines of code) as it is already supported by Mono.CSharp.  However, multi-line support for loose function will require some work.  So do we want to split this into 2 PRs or wait for a full fix?

adamralph:
@ztone thanks for the info!

I would prefer splitting it up. If you can send a separate PR now for 'multi-line support for everything but loose functions' then we can get that in now and perhaps even make it part of 0.13. The loose function support can come later and we can spin off a separate issue for it.

adamralph:
Fixed in #771

ztone:
[...]

PR #771, opened 2014-06-23
Mono engine using code segments to better parse scripts

### LINKED ISSUE #664: Mono error reporting
Currently the Mono engine uses the built in Mono.CSharp console printer directly.

Should instead implement a custom printer that will capture errors & warnings and allow scriptcs infra to handle them

### LINKED ISSUE #713: Expressions capturing local vars do not compile in Mono
```c#
var x = 123;
Action a = () => x++;
Console.WriteLine(x);
```

image

ztone:
This PR is a proposed solution to get the Mono engine close to be at par with Roslyn and solve most of the Mono issues.

* Fixes #713 (capture local vars)
* Blocks in script no longer interferes with parsing (had an commented out test for this in SyntaxParserTests + had this problem when investigating #740)
* Fixes #664 (Mono error reporting)
* Returns compile errors in ScriptResult from Mono.CSharp and NRefactory (when rewriting methods) and displays the correct line of the erroneous code (not supported before).
* Partly addresses #665 (does still not support loose functions in multi-line)

**Why this approach**
NRefactory failed to parse correctly if script contained blocks <code>{ }</code>
Needed a way to fix #713

**What this PR does**
So the approach I've taken is to split the script into code segments.

* The lexer : find a few token needed for parsing and disregard stuff in strings, chars and comments so it doesn't interfere with the parsing.

* Region parser: Eats all tokens until it find either a semicolon or an opening/closed bracket.

For example <code>Console.WriteLine(“hello”);</code> is a region and <code>if(true) { doStuff(); }</code> is also a region.  The lexer and parser don't care about the syntax or semantics, it's just used to do a logical split of the code.

* NRefactory: Each segment is then passed into the NRefactory parser to check if the segment is a class, method or evaluation.  It the segment is a method it is rewritten by NRefactory to a method expression to support loose functions.

Segments are also ordered as we want to evaluate classes first, then method prototypes.  After that we evaluate method expressions and lastly the evaluation.

If you look at ScriptCs code this approach does make some sense.
```c#
class A { you can write anything here }

void Bar() { also here }

Bar( even inside parentheses );
```
Here we identify three regions/segments.  The Mono.CSharp evaluator will handle all the compiler errors that are inside each segment (and NRefactory for methods).

**Other**
* Mono engine will return the correct line number for a compilation errors.  However, a few method errors are still a bit strange because of the method rewrite.  Needs investigating.

* Before, code like <code>error.</code> didn't cause any failure (as the evaluator regards this as a partial, because of the dot).  Handling partial loose functions is going to be a bit tricker and needs to be in another PR.

Please test drive.

Stone


ztone:
@filipw @adamralph  I rebased this PR.  

Once I took a closer look I remembered that this PR has some needed features like keeping track of line numbers (as scripts are being re-ordered before evaluation) and maintains code format as classes and methods are re-formatted by NRefactory in our current solution.  This is needed when we want to add our custom mono reporting class.  Also because of this reformatting, implementing multiline support for classes is going to be awkward in our current solution. For example, NRefactory will add a missing curly bracket for classes which will mean that the Mono.CSharp will not think it’s an incomplete submission.

**Pros:**
* Fixes block parsing errors
* Fixes #713 as the newest Mono.CSharp does not have that fix yet.
* Keeps track of line-number (to be used with the mono reporter)
* Keep code format intact so it will be easier to add multi-line support for classes.

**Cons:**
* This is a large chuck of code that might contain unforeseen parsing bugs, but is has a great deal of unit-tests.
* A small performance hit as script needs to be pre-parsed.

filipw:
great. IMHO this is good to merge. I know this is a massive changeset, but it was expected to be, since it introduces a lexer. I wouldn't worry about introducing bugs that much. Firstly, it has a bunch of tests like you mentioned. Secondly, I tested it out when it was brought forward originally and it worked like a charm.

I say this should go in immediately, as it's definitely an improvement in our Mono support. unless someone wants to chime in? @scriptcs/core

khellang:
:shipit:

adamralph:
Thanks @ztone! This will go out in 0.13.

Issue #175, opened 2013-04-05
Consider refactoring our arg design to use sub commands


glennblock:
Per Adam

We should consider not just using flat args, but actually having sub commands similar to git.

https://github.com/adamabdelhamed/PowerArgs#advanced-example

adamralph:
(From https://github.com/scriptcs/scriptcs/pull/868#issuecomment-68596167)

Current command line args:

### Commands
* run a script: implicit by presence of script name
* start REPL: implicit by absence of script name or `-repl` when other args are present (this is confusing but is addressed by https://github.com/scriptcs/scriptcs/issues/869)
* show help `-help`
* install package `-install`
* save packages.config file `-save`
* clean installed packages `-clean`
* show version `-version`

### Options
* drop into REPL after running script file `-repl`
* emit PDB symbols `-debug`
* assembly in memory or filesystem `-cache`
* log level `-loglevel`
* install package globally `-global`
* allow pre-release packages `-allowprerelease`
* watch script file for changes `-watch`
* module to load `-modules`

One thing to note is that if we want to maintain support for `scriptcs foo` to run a script named `foo.csx` then we would have to exclude certain script names from this behaviour, i.e. `help.csx`, `install.csx`, `save.csx`, `clean.csx` and `version.csx`. I guess for these cases we would have to insist on the extension being supplied. Also note that if we ever introduce new commands, we will have to make a *breaking change* to exclude a script with that command name from the extensionless behaviour.

An alternative approach is to move `install`, `save` and `clean` over to a new `spm.exe` and leave just `-help` and `-version` as commands disguised as options in `scriptcs.exe`. This would match the situation with node and npm.

PR #868, opened 2014-12-11
prefixed FileSystem locations with scriptcs_

### LINKED ISSUE #626: Change scriptcs filesystem names to avoid clashes
Currently scriptcs uses `/packages`, `/bin` and `packages.config`. These are susceptible to name clashes in some scenarios, e.g. a scriptcs based build system such as Bau or Shovel will typically have a .csx in the solution root so the `/packages` folder for the solution will clash with the `/packages` folder for scriptcs. Another scenario is descibed in https://github.com/scriptcs/scriptcs/issues/618. It's also common to use `/bin` for other purposes, e.g. in my day job, all builds write their resulting artifacts into `/bin`. Regarding `packages.config`, it's not unreasonable to imagine scenarios where someone might want to do some scripting within a .NET project folder, so there would be a name clash with the project's `packages.config`.

nodejs uses `/node_modules`, which almost certainly guarantees that nodejs can operate in any scenario, isolated from the file system environment in which scripts are located. Following that pattern, and to avoid clashes with a .NET projects `packages.config`, I propose that scriptcs changes to using `/scriptcs_packages`, `/scriptcs_bin` and `scriptcs_packages.config`.

There are other issues for allowing custom script pack and module paths (#618 and #625). If they are implemented then `/scriptcs_packages` and `/scriptcs_bin` would simply be the defaults, instead of `/packages` and `/bin`.

I concede that the proposed names are a little long and slightly ugly and I almost feel reluctant to suggest them but ultimately if we want to allow people to blend together the two ecosystems (regular .NET and scriptcs) in weird and wonderful ways then I think this change is necessary. Alternatively if we do implement #618 then we could leave it to CLI tool authors to document where script packs should be installed but this would add some friction for users of those tools and it wouldn't solve the `packages.config` clash. At the moment I'm keen to explore if we can offer an experience closer to nodejs for tools like gulp and grunt, where things like this don't need thinking about by users.

This will prove disruptive in the sort term since current users will have to do some renaming (we could provide a `-migrate` CLI switch) but I think it will remove a lot of friction in the long term in supporting build scripting and other scenarios. If people agree then the sooner this change is made before 1.0, the better.

adamralph:
fixes #626

### DIFF
--- a/src/ScriptCs.Contracts/IFileSystem.cs
+++ b/src/ScriptCs.Contracts/IFileSystem.cs
@@ -38,6 +38,8 @@ namespace ScriptCs.Contracts
 
         void Move(string source, string dest);
 
+        void MoveFolder(string source, string dest);
+
         bool FileExists(string path);
 
         void FileDelete(string path);
--- a/src/ScriptCs.Core/FileSystem.cs
+++ b/src/ScriptCs.Core/FileSystem.cs
@@ -159,27 +159,27 @@ namespace ScriptCs
 
         public virtual string BinFolder
         {
-            get { return "bin"; }
+            get { return "scriptcs_bin"; }
         }
 
         public virtual string DllCacheFolder
         {
-            get { return ".cache"; }
+            get { return ".scriptcs_cache"; }
         }
 
         public virtual string PackagesFile
         {
-            get { return "packages.config"; }
+            get { return "scriptcs_packages.config"; }
         }
 
         public virtual string PackagesFolder
         {
-            get { return "packages"; }
+            get { return "scriptcs_packages"; }
         }
 
         public virtual string NugetFile
         {
-            get { return "nuget.config"; }
+            get { return "scriptcs_nuget.config"; }
         }
 
         public virtual string GlobalOptsFile
--- a/src/ScriptCs.Core/FileSystem.cs
+++ b/src/ScriptCs.Core/FileSystem.cs
@@ -83,6 +83,11 @@ namespace ScriptCs
             File.Move(source, dest);
         }
 
+        public virtual void MoveFolder(string source, string dest)
+        {
+            Directory.Move(source, dest);
+        }
+
         public virtual bool FileExists(string path)
         {
             return File.Exists(path);
--- a/src/ScriptCs/Command/CommandFactory.cs
+++ b/src/ScriptCs/Command/CommandFactory.cs
@@ -143,6 +143,11 @@ namespace ScriptCs.Command
                 return new CompositeCommand(installCommand, new SaveCommand(packageAssemblyResolver, _fileSystem, logger));
             }
 
+            if (args.Migrate)
+            {
+                return new MigrateCommand(_fileSystem, logger);
+            }
+
             // NOTE (adamralph): no script name or command so assume REPL
             var scriptServices = _scriptServicesBuilder.Build();
             var replCommand = new ExecuteReplCommand(
--- a/src/ScriptCs/Command/ICommand.cs
+++ b/src/ScriptCs/Command/ICommand.cs
@@ -40,6 +40,10 @@ namespace ScriptCs.Command
     {
     }
 
+    public interface IMigrateCommand : ICommand
+    {
+    }
+
     public interface IDeferredCreationCommand<TCommand> : ICommand where TCommand : ICommand
     {
     }
--- a/src/ScriptCs/Command/MigrateCommand.cs
+++ b/src/ScriptCs/Command/MigrateCommand.cs
@@ -1,0 +1,61 @@
+﻿using System.Collections.Generic;
+using System.Globalization;
+using Common.Logging;
+using ScriptCs.Contracts;
+
+namespace ScriptCs.Command
+{
+    internal class MigrateCommand : IMigrateCommand
+    {
+        private readonly IFileSystem _fileSystem;
+        private readonly ILog _logger;
+
+        public MigrateCommand(IFileSystem fileSystem, ILog logger)
+        {
+            _fileSystem = fileSystem;
+            _logger = logger;
+        }
+
+        public CommandResult Execute()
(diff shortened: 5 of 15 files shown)


adamralph:
@scriptcs/core any other opinions on `migrate` as command or a module?

If a module, then we may need to do something about the UX.

Currently our command line args are mix of commands disguised as options and actual options (we should really consider implementing https://github.com/scriptcs/scriptcs/issues/175) :

### Commands
* run a script: implicit by presence of script name
* start REPL: implicit by absence of script name or `-repl` when other args are present (this is confusing but is addressed by https://github.com/scriptcs/scriptcs/issues/869 which is currently also solved by this PR)
* show help `-help`
* install package `-install`
* save packages.config file `-save`
* clean installed packages `-clean`
* show version `-version`

### Options
* drop into REPL after running script file `-repl`
* emit PDB symbols `-debug`
* assembly in memory or filesystem `-cache`
* log level `-loglevel`
* install package globally `-global`
* allow pre-release packages `-allowprerelease`
* watch script file for changes `-watch`
* module to load `-modules`

With the current design, the option to load the `migrate` module is just that, an option. The command to be performed will have to come from one of the listed commands above. With the change to address #869, typing `scriptcs -m migrate` will work, but will drop the user into the REPL afterwards. Typing `scriptcs foo.csx -m migrate` will also work but will obviously also run the script. There is no way to "just migrate", which I'm quite sure I would want as a user so I could assess the outcome before going ahead and trying to execute scripts or REPL. Of course, someone could type something like ``scriptcs -m migrate -version` but that's just weird.

adamralph:
OK, I decided to make migration a command. There were just too many outstanding issues and questions around 'modules as commands' and making it a real command turned out to be far more straightforward and less disruptive. Yes, we may take away the migrate command somewhere down the line but I don't consider this a problem.

This PR is far more straightforward now and the commit stream should be self explanatory. It also includes a full acceptance test for migration. The fix for #869 is no longer included here since it is irrelevant. I will send a separate PR for that issue.

adamralph:
Builds are green. This PR is good to go again.

khellang:
Needs a rebase :smile

adamralph:
Done :smiley:

khellang:
How will people know that they need to do this once their scripts break? Docs? Could we detect and run the migration if needed? Or is it too hard to detect? I.e. what if the user doesn't want to migrate existing packages.config etc.?

adamralph:
I think we need to shout about it quite loudly, perhaps a post on http://blog.scriptcs.net/ and then we tweet it around etc, and/or cross post on other blogs owned by @scriptcs/core members.

Detection isn't really possible since, in a given folder, we never know whether `/packages`, `packages.config`, etc. is meant for scriptcs or regular .net projects.

We need to make it clear in our announcements exactly what `scriptcs -migrate` will do. In the case of an existing directory where they have a packages.config which they don't want to migrate then I can't see there being a need to do any migration in that directory at all.

khellang:
Alright :shipit:

glennblock:
Hi @adamralph 

@khellang and I were just talking about this and we both share the concern about how this will break customers. I am worried also about the person who falls in between the cracks where the current migrate won't help them. 

Imagine I ran scriptcs within my full .NET app folder. Now you are migrating and moving my core app packages.config to the new scriptcs one. That may make scriptcs work but it breaks my .NET app as the packages.config is gone.

Even if migrate worked, it is still ugly and non-intuitive as the user won't get a warning that would lead them to know to do migrate.

So here's my suggestion. 

1. We get rid of the migrate command.
2. We automatically detect if there is a packages.config and packages present. If there is then we COPY packages.config and packages to the new folder.
3. We give the user a clear message that details the upgrade.

By doing a copy instead of a move, we don't affect their existing app. Also going forward if they install new packages with scriptcs-install they will automatically go to the right place.

The downside of this approach is the file copy will result in pulling in packages that may not be needed. One can always go manually and clean those out if they care.

Thoughts?

adamralph:
I really like this idea and I think it's better than what we currently have. I'll send a new PR which reverts the last two commits from this PR and adds what you've described.

What about the other artifacts? I think we'll have to do the same for `bin/` and `nuget.config`. I'm not sure about `.cache/`.

glennblock:
@adamralph great.

For .cache it is safe to rename I think.
For nuget.config, copy if we have our own.
For bin, probably safest to copy just in case.

adamralph:
One remaining question is where to put this code? I can very easily put it in scriptcs.exe, but what about hosting scenarios? Do we want to offer the ability for hosts to perform the same operation? If so, I'm not really sure where it would need to go.

adamralph:
Also, if we do want to expose it for hosting, should it be implicit for anyone using the hosting layer or should it require an explicit invocation?

adamralph:
I have most of the work done and I'll send a PR soon.

The approach I have gone with as follows:

I introduced first class types for migration, i.e. `ScriptCs.Contracts.IFileSystemMigrator` and `ScriptCs.FileSystemMigrator` (in ScriptCs.Core.dll) and given them the full wire up, i.e. in `ScriptServices`, etc.

`FileSystemMigrator` first looks for the existence of any of the new file system artifacts. If it detects at least one artifact then it assumes migration (either automatic or manual and either full or partial) has already been done and it does not attempt any migration. If no new file artifacts are detected then any legacy file system artifacts are migrated, using the logic described in the above comment by @glennblock.

Each `ICommand` implementation that needs to do migration before execution, receives an instance of `IFileSystemMigrator` which it invokes as soon as `Execute()` is called.

The commands which perform migration are: script execution, REPL, `-install`, `-save` and `-clean`. The commands which do not perform migration are `-help` and `-version`.

Before sending the PR, I'm going to add acceptance tests for `-install`, `-save` and `-clean` to make sure the change doesn't break them.

If anyone has any comments on the approach, now is the time :wink:.

PR #887, opened 2015-01-03
Ensure ArgumentParser does not return null args

adamralph:
Addresses https://scan8.coverity.com:8443/reports.htm#v21582/p10801/fileInstanceId=3240414&defectInstanceId=1396800&mergedDefectId=96015


khellang:
Friends don't let friends return null :wink:

PR #890, opened 2015-01-04
changed command line to assume REPL when no command or script name is passed

### LINKED ISSUE #869: Default to REPL when other options are passed
For a bare command with no options, i.e. `scriptcs`, the presence of a script name determines whether to enter the REPL (with the `-repl` option forcing exit to REPL even when there *is* a script name).

However, as soon as any other option is provided, e.g. `-modules "mono"` or `-debug`, the `-repl` option *must* be provided otherwise invalid usage is assumed and the usage is shown in the console. IMO this is nasty UX. It bites me *every single time* I try to run the REPL with the mono engine :tired_face:.

adamralph:
fixed #869

### DIFF
--- a/src/ScriptCs/Command/CommandFactory.cs
+++ b/src/ScriptCs/Command/CommandFactory.cs
@@ -59,18 +59,18 @@ namespace ScriptCs.Command
 
             if (args.Repl)
             {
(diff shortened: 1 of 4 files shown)


khellang:
Yeah, this is a better default :smile:

Issue #893, opened 2015-01-05
AssemblyResolver should only enumerate top level for bin


glennblock:
Currently the resolver is greedy when it finds assemblies and will look in all child folders underneath the bin folder. This is problematic in custom hosting scenarios where packages may be under bin, as it will pick up all assemblies of all targets.

PR #894, opened 2015-01-05
Fixes AssemblyResolver.GetBinAssemblyPaths to not find assemblies in chi...

glennblock:
...ld folders under bin

Fix for #893


khellang:
:heart:

adamralph:
Parent issue is already labelled as a bug, so we don't want to double count.

Issue #884, opened 2014-12-25
0.13 release


adamralph:
- [x] close https://github.com/scriptcs/scriptcs/issues/626 (@adamralph)
- [x] close https://github.com/scriptcs/scriptcs/issues/869 (@adamralph)
- [x] ensure all 0.13 issues are attached to the milestone and labelled appropriately (@adamralph)
- [x] bump version number - PR raised https://github.com/scriptcs/scriptcs/pull/905 (@adamralph)
- [x] draft GitHub release (@adamralph)
- [x] raise PR from dev to master (@adamralph) - https://github.com/scriptcs/scriptcs/pull/907
- [x] raise PR for blog post regarding file system changes in #626 (@adamralph) - https://github.com/scriptcs/scriptcs.github.com/pull/1
- [x] merge PR from dev to master (@adamralph)
- [x] merge PR for blog post (@glennblock)
- [x] publish GitHub release (@adamralph)
- [x] announce

adamralph:
Needs https://github.com/scriptcs/scriptcs/issues/626 to be closed (currently half implemented in dev).

khellang:
Yay! :+1:

PR #902, opened 2015-01-16
automatic filesystem migration

### LINKED ISSUE #626: Change scriptcs filesystem names to avoid clashes

adamralph:
closes #626


khellang:
I think this looks nice. Had a couple of questions, but I think it's good to go. @adamralph Tell me if I should go ahead and merge anyway :smile:

adamralph:
@khellang thanks for reviewing.

With regard to the calls to `Build()`, I don't mind them as they are, but if you want, I can make the refactoring I suggested in my reply. In fact, it probably makes sense to check `args.Version` first since that really has "don't run any command just show me the version" semantics.

As for the constants in the tests, I'm not huge fan of that. Constants are a mechanism for declaring something which is logically eternally constant so I would't make them public, and a private constant wouldn't achieve anything, so I'd prefer to leave as is.

khellang:
Time to close #884?

adamralph:
Absolutely!

Issue #874, opened 2014-12-18
Cannot install ScriptCS via Chocolatey for multiple user accounts.


aarondandy:
After ScriptCS is downloaded to `C:\ProgramData\chocolatey\lib\ScriptCs...` I think `chocolateyInstall.ps1` is run which copies the files to the `LOCALAPPDATA\scriptcs` folder for the current user. This works for the first account but when I log into the system as another user and attempt to install for that user it fails because ScriptCS is already installed. Because of how our security here is set up I can't elevate my account but instead run programs as a different accounts with different privileges. I know this is an inconvenient setup but my employer has lots of computers so I don't think I am unique in this situation.

As a short term (or possibly "short term") work around would it be safe for me to add `C:\ProgramData\chocolatey\lib\ScriptCs.{version}\tools\scriptcs` to the system path?

tillig:
Just ran into this myself. Dos there need to be any user-specific portion to the install? If scriptcs doesn't care where it's run from, could it use the built-in Chocolatey mechanism to add a batch redirect to the `chocolatey/lib/scriptcs.*/scriptcs.exe`?

I could be over-simplifying here, I just see stuff like the `putty.portable` package able to put in a near-zero install and magic happens.

adamralph:
@scriptcs/core can anyone chime on this? The choco package was put together before my time.

I think the current behaviour may have been driven by the desire for the installer to not require admin rights.

aarondandy:
With an MSI it can detect if you are elevated (or ask) and install to the appropriate location. Also, the "this user" or "all users" question can be asked. But that is going down the WiX road. On a related note @dealproc mentioned in JabbR that they have experience with WiX ;)

tillig:
I found a solution to this (PR #903) by modifying the Chocolatey package contents so the new Chocolatey package should enable a single install on the machine to work for multiple users. No WiX required.

aarondandy:
I think that PR closes this issue for sure and will satisfy the majority of people. I still think there will be a small group that for one reason or another will prefer an msi file but I wont push that as I don't think the complexity is worth it. Perhaps a zip distribution would satisfy them as an alternative, should that issue be raised in the future.

khellang:
This should be resolved with #903 :smile:

PR #903, opened 2015-01-19
Chocolatey installation for multiple users

tillig:
Fix for issue #874. Switched the installation script to use automatic Chocolatey facilities for installing command-line tools. Chocolatey handles putting in a shim to scriptcs.exe and adding the path info. No need for uninstall when using this mechanism - Chocolatey does it all for you.


tillig:
Testing this is sort of painful and manual, but here's how:

1. Build the solution so the Chocolatey package gets created.
1. Create a temporary folder `C:\temp\temp-scriptcs`.
1. Copy the Chocolatey package `scriptcs.0.12.0.nupkg` into the temp folder.
1. Rename the Chocolatey package from `scriptcs.0.12.0.nupkg` to `scriptcs.1.0.0.zip`. Rename to zip so you can use 7-zip (or other tool) to unpack it; update the number so Chocolatey will eventually think it's more recent than previously released packages.
1. Unpack the zip file into a folder `scriptcs.1.0.0`.
1. Delete the old `scriptcs.1.0.0.zip`.
1. Update the .nuspec in there to say version 1.0.0.
1. Zip up the contents of the folder into a new version of `scriptcs.1.0.0.zip`. Make sure if you browse in the zip file that the .nuspec shows up right in the root of the zip - if you see a folder in there `scriptcs.1.0.0` then you zipped at the wrong level.
1. Rename the .zip file back to `scriptcs.1.0.0.nupkg`.
1. Open up a Powershell script as administrator.
1. Run `choco sources add -name temp-scriptcs -source C:\temp\temp-scriptcs`
1. Run `choco install scriptcs -version "1.0.0"`
1. Watch as the Chocolatey installer runs.
1. Without closing down the Powershell prompt, try using `scriptcs.exe` - it should work.
1. **Open a separate Powershell prompt as a non-admin user (or an otherwise different user than was used to install the package) and run `scriptcs.exe` - it should work for the other user, too.** (This is the bit that got fixed by the PR here.)
1. Browse to `C:\ProgramData\chocolatey\bin` and you'll see a `scriptcs.exe` shim in there.
1. Back at the admin prompt, run `choco uninstall scriptcs`.
1. In `C:\ProgramData\chocolatey\bin` the `scriptcs.exe` shim should be gone.

adamralph:
@tillig thanks very much for this! I really like it, especially the hugely simplified install script! The use of built in chocolatey idioms is a way nicer way to go.

I guess the only downside to this is that it will now require admin rights to install scriptcs. @scriptcs/core how much does this bother us?

aarondandy:
Thank you very much! :+1:

tillig:
Generally packages installed by Chocolatey do seem to require admin rights. It's totally up to y'all, but there's at least precedent for that. (Technically, for me, at least, even installing scriptcs in its current incarnation required admin rights because once you install a package using admin rights, the admin user takes ownership of the chocolatey.log and further non-admin installs become impossible.)

Anyway, if this looks like it'll work and folks are amenable, it'd be pretty helpful for those of us switching between admin and non-admin accounts.

aarondandy:
I get dirty messages from Chocolatey if I don't install things as my other user account with admin privileges. Sometimes, those messages seem accurate.

glennblock:
Nice! I am not too concerned on the admin requirement for chocolatey as it
seems like running as admin is not unreasonable.

adamralph:
@glennblock great - in that case I think we should go ahead and accept this.

@tillig one question first:- how does it behave if an older per-user installation is already present?

tillig:
Just tested it out and apparently [upgrading a package does not actually run the uninstall for the previous version](https://github.com/chocolatey/chocolatey/issues/291) so the way I have it now you'd have to manually uninstall and then reinstall.

Let me see if I can make it just a touch smarter so the upgrade scenario works.

tillig:
I added a block of code that will remove the old %AppData% based install locations and per-user path settings during an upgrade scenario. This allows existing users to upgrade to the new mechanism and new users to install.

It's unfortunate, but [there's no Uninstall-ChocolateyPath](https://github.com/chocolatey/chocolatey/issues/97) in Chocolatey so we can't just use a one-liner to remove the path like was used to install it. Thus you'll see a little bit of manual parsing, but there's really no other way. It's the reverse logic of what's in `Install-ChocolateyPath`.

khellang:
Thanks @tillig! :+1:

khellang:
Added you to the contributors list: https://github.com/scriptcs/scriptcs/wiki/Contributors :smile:

khellang:
BTW, @scriptcs/owners @scriptcs/core: we should really automate that list. Maybe I should add a list as part of the new community blog? Like http://nancyfx.org/contribs.html

glennblock:
@khellang We could host it on the blog and link to it.

adamralph:
It's actually already automated at https://github.com/scriptcs/scriptcs/graphs/contributors

It does require that people commit with their correct email addresses though.

tillig:
Happy to help!

ztone:
Damn, I've been struggling to get on that wiki list for years ... what does a man have to do?

adamralph:
:scream: you weren't on it!?

Fixed :smiley:

filipw:
It's a wiki, it's publicly editable :-)

ztone:
Are you saying that I could have just added myself a long time ago and didn't have to learn to write any code at all. Man, that's several years down the drain.

ztone:
Look ma .. on top of the list 
https://github.com/scriptcs/scriptcs/wiki/Contributors

(sorry for hijacking this thread for such silliness)

adamralph:
@khellang I like your idea of automating the list in a similar way to Nancy and I think you should go for it.

It will be a repitition of https://github.com/scriptcs/scriptcs/graphs/contributors but in a nicer looking and more prominent way.

adamralph:
https://github.com/scriptcs/scriptcs/issues/904

PR #920, opened 2015-02-05
fix for 700: changed ObjectSerializer to remove $id and $ref unless there is a circular ref

### LINKED ISSUE #700: Remove redundant $id from JSON strings in REPL
$id and $ref are required. This is a given. When serialising an object graph you need them to deal with circular references. The only way to deal with a circular reference is to add a $ref pointing to the $id of the circular ref'd object.

However, we can remove all instances of $id which do not have a corresponding $ref, since they are redundant. From https://github.com/scriptcs/scriptcs/pull/696#issuecomment-43481416

> IMO you should serialize directly to LINQ to JSON objects, navigate through the object tree and remove the $id properties that don't have a corresponding $ref, then ToString() to get the JSON and print it.

adamralph:
fixes #700

### DIFF
--- a/test/ScriptCs.Hosting.Tests/ObjectSerializerTests.cs
+++ b/test/ScriptCs.Hosting.Tests/ObjectSerializerTests.cs
@@ -1,5 +1,5 @@
 ﻿using System;
-
+using Should;
 using Xunit;
 
 namespace ScriptCs.Hosting.Tests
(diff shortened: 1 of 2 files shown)


test/ScriptCs.Hosting.Tests/ObjectSerializerTests.cs
```diff
@@ -16,6 +16,34 @@ public TheSerializeMethod()
...
+  ""Bar"": {
+    ""Baz"": true,
+    ""Bazz"": 123.4,
+    ""Bazzz"": ""hello""
+  }
+}");
+            }
+
+            [Fact]
             public void ShouldSerializeTypeMethods()
```
glennblock: Would be nice to change this test and ShouldSerializeDelegates to verify the result similar to what you did in ShouldSerialize. All it does now is verify that there is no exception, not that the result is correct ;-)

glennblock:
Overall LGTM, but I think we should improve the two tests I mentioned if we are in this code.

adamralph:
Very good point. I'll add that shortly.

adamralph:
@glennblock the serialization of `Type` methods is *enormous*. Are we sure we want to do this? IIRC this test was added after I witnessed `typeof(Type).GetMethods();` blow up on stage for Jon Skeet at NDC London and ensuring an exception wasn't thrown was it's specific purpose.

The serialization of the `Action` and `Func` is of a manageable size.

adamralph:
Ugh... there's also a problem with asserting the `Action` and `Func`. The names of the hoisted methods are not deterministic. E.g. if I run just the single test method, the hoisted method is named `<ShouldSerializeDelegates>b__3`. If I run the entire test class, the hoisted method is named `<ShouldSerializeDelegates>b__6`.

I think we'll have to leave these tests as they are, but I've refactored them to 3A's, instead of using the `Assert.Throws()`.

adamralph:
Ultimately, we're not trying to assert the serialization of all object types in this test class, but there are specific cases for which we originally had errors, i.e. the methods of `Type` and any `Action` or `Func`, so at least those tests are locking in the serialization of those types without errors.

glennblock:
OK, sounds good.

PR #924, opened 2015-02-07
Ensuring GAC dlls are loaded. Fix for #922

### LINKED ISSUE #922: scriptcs 0.13.1 crashes when a package has a framework assembly reference
If the packages folder contains a package which has a framework assembly reference, e.g. ScriptCs.Contracts, which references `System.ComponentModel.Composition`, then scriptcs.exe crashes and dies.

This is because, as part of the fix for #913, it's now passing all the assemblies through `AssemblyUtility.IsManagedAssembly()`, which in turn calls `AssemblyName.GetAssemblyName(path)`, but `"System.ComponentModel.Composition"` is not a path, so a `FileNotFoundException` is thrown, and unhandled.

`AssemblyUtility.IsManagedAssembly()` is already catching `BadImageException` and returning false in that case, so I guess it should also catch `FileNotFoundException` and also return false in that case.

Note, that the script pack acceptance test is passing, even with this bug present, because ScriptCs.Adder doesn't take a dependency on ScriptCs.Contracts but ships the DLL in the package instead.

I thought perhaps we'd like to try a 'patch master' approach for this (so dev is free to progress independently in the meantime):

- [x] draft GitHub release (@adamralph)
- [x] raise PR to master (including version bump) (@glennblock and @adamralph)
- [x] merge PR to master (@adamralph)
- [x] publish GitHub release (@adamralph)
- [x] announce (@adamralph)
- [x] merge master to dev (@adamralph)

adamralph:
fixes #922
supsersedes #923

### DIFF
--- a/build/ScriptCs.Version.props
+++ b/build/ScriptCs.Version.props
@@ -5,7 +5,7 @@
   <PropertyGroup>
     <MajorVersion>0</MajorVersion>
     <MinorVersion>13</MinorVersion>
-    <PatchVersion>1</PatchVersion>
+    <PatchVersion>2</PatchVersion>
 
     <!-- Change this to set the build quality of the project. Use values like "alpha", "beta", "rc1", "rtm", etc. -->
     <!-- These values are used in SemVer, so make sure to always increase these alphabetically. -->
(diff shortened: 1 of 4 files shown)


glennblock:
Great!

PR #926, opened 2015-02-07
923 refactoring

### LINKED ISSUE #922: scriptcs 0.13.1 crashes when a package has a framework assembly reference

adamralph:
This is a refactoring of the changes in #923 which fixed #922.

Whilst that PR fixed the bug and allowed us to ship 0.13.2, I was quite uncomfortable leaving `AssemblyUtility.IsManagedAssembly()` returning `true` whenever the the path passed to it isn't rooted and doesn't end in 'dll' or 'exe'. For instance, I could pass `"ThisDoesNotExist.txt"` into it and it would return `true`.

In this PR I've reverted `AssemblyUtility` and instead changed `PackageAssemblyResolver` to filter out unmanaged assemblies from the list returned from `IPackageObject.GetCompatibleDlls()` before concatenating them with `IPackageObject.FrameworkAssemblies`.

IMO the quickest way to review this PR is one commit at a time.


glennblock:
While true, it is not really an issue as the classes that use
AssemblyUtility only scan dlls and exes. The text file will never get
loaded or indicated as a reference.

I don't have a big concern but I don't think was needed.

adamralph:
It's true that the original fix was fine given all the calls made to `AssemblyUtility.IsManagedAssembly` from the scriptcs codebase, but the method is public and packaged in core, so I think it's important that it conveys the correct semantics. In fact, even if it was internal, I'd still consider this important.

adamralph:
@scriptcs/core does anyone fancy merging this? Are there any concerns? It should be quite quick to review if you take one commit at a time.

khellang:
@adamralph :+1:

adamralph:
woot! thanks @khellang

Issue #238, opened 2013-05-08
Introduce script packs as script.


glennblock:
Yes we're going there. This will enable to you share scripts in nuget packages as script modules.

It will work very similar to node. Module scriptcs can depend on other modules. All modules will get installed in a modules folder.

Yes, we do love the node module system!

glennblock:
This is still in progress. I have rethought this and am taking a different route now which will be more efficient, remove any extra load-time impact and be cleaner. I hope to make progress in the next month or so.

adamralph:
Dropped in favour of https://github.com/scriptcs/scriptcs/issues/909

Issue #909, opened 2015-01-29
Script Libraries


glennblock:
This feature will enable scriptcs to load up scripts embedded within NuGet packages. This is not the same as scripted script-packs.

adamralph:
closed by #928

Issue #767, opened 2014-06-18
Nested script dependencies


adamralph:
_Dependent upon #238 or #909, whichever is favoured_

Currently scriptcs uses a flat dependency model. I.e. under packages, the dependencies are contained in a flat folder structure and only one version of each is loaded at runtime.

This contrasts with nested dependencies, as demonstrated by node and K, where each dependency has it's own nested copy of each of it's dependencies.

scriptcs is primarily constrained to flat dependencies due to CLR limitations with assembly loading. (scriptcs running on K, on the hand, will automatically benefit from K's nested dependencies.)

With the introduction script based packages, I believe we have the opportunity to introduce nested dependencies, since each dependency is just a script rather than an assembly.

PR #928, opened 2015-02-11
Introduces the ability to author, package, and consume Script Libraries 

### LINKED ISSUE #935: Move adding of ScriptCs.Core reference from Roslyn ScriptEngine to ScriptExecutor
Currently the Roslyn Engine is adding a reference for ScriptCs.Core, but the Mono engine does not. This was causing the PackageScripts feature acceptance tests to fail on Mono. This fix will ensure all engines have a reference to ScriptCs.Core.

glennblock:
... up in NuGet packages

_The future is here_

This addresses #909 and #238. It also fixes #935 which was causing ScriptLibrary acceptance tests on TeamCity / Mono to fail.

With this PR, you can author a Script Library as a set of csx files which you then package up in a NuGet package. The packages are kind of the scriptcs equivalent of source packages. You can then install the package and use it within a script. 

Within the package, the scripts must be in a content/scriptcs folder. Each script must also be added to the nuspec file as a file entry. The packs are not accessed via Require<T>, instead they are simply newed up.

For an example showing all the files in a Script Library, see this [gist] (https://gist.github.com/glennblock/c39cd3ed98ecb4e2d020)

Notice it depends on another script pack, `ScriptCs.Logger` which is required. 

Once the package is created i.e. `ScriptCs.Calculator.0.1.0.nupkg` and placed one of your NuGet sources it can be installed with `scriptcs install`. 

Here is client code that can then use the Script Library:

```csharp
var calc = new Calculator();
var result = calc.Add(5,2);
Console.WriteLine(result);
Console.ReadLine();
```

A few constraints / notes:

* You cannot have loose code in your package script. It must be contained within a method. This is because your script is packaged up inside of a wrapper class.
* You must have an entry level script with the format of XXXMain.csx i.e. CalculatorMain.csx in the example.
* You can use Require<T>. To do so create an instance field and then use Require. Require<T> is a static method in this context so it will work.
* You can use #load and #r. For example in the gist you can see I have `#load Multiply.csx` which loads up the file which contains the Multiply method and includes it. 

You can read more in the design doc [here] (https://github.com/scriptcs/scriptcs/wiki/Script-Libraries)


glennblock:
(releases breath)

glennblock:
A picture tells a thousand words, so here's a package script (scripted script pack) running in the REPL.

screen shot 2015-02-11 at 9 45 18 pm

glennblock:
Not sure why Travis is failing. It fails on a few new acceptance tests, though they appear to work on TC?

adamralph:
Great work @glennblock. I really like the approach of the entry level script and 'main' class exposing the top level methods. From there, the script author can more or less provide anything they like, e.g. other types through factory methods, accessor properties, etc. and the sky's the limit!

I haven't had a chance to dig into the implementation yet, but I have a few questions from a functional perspective:

1. You mention that it is still a work in progress. What is left to be done? I.e. if we were to merge now, would it be releasable?
1. You mention that it closes both #909 and #238. It seems as though #909 supersedes #238 favouring 'scripts as packages' (or whatever we end up calling them) over 'script packs as script'. If so, I think we should close #238 as a `wontfix`.
1. I agree that we need to be careful with the naming. Perhaps, as part of avoiding confusion with 'script packs', we should consider a different naming convention for the entry level script than 'XXXPack.csx'? Perhaps 'XXXMain.csx'?
1. How does this work when another language engine is used at runtime? E.g. if I run using the F# engine, are the content/scriptcs scripts simply ignored due to their file extensions?

I'd also like to open up a conversation about the merged script caching at installation time which, whilst ostensibly an implementation detail, I think may have significant functional implications. It may indeed be the right solution but I think it's important to consider the implications, ensure we are comfortable with it and that we are doing it for the right reasons. The more of @scriptcs/core that can chime on this, the better.

Firstly, is it proven that the caching is required? Is this purely a performance optimisation? If so, how long does the script loading and merging take? If it moved from installation time to runtime what would be the performance hit? Other platforms do indeed read their dependencies from the filesystem at runtime with acceptable performance so I'm interested to know how scriptcs would fair with that approach. (Note that if merging happened at runtime then writing the merged script to disk would serve no purpose, apart from perhaps as a debugging aid when `-debug` is passed.)

My specific concerns stem mainly from the introduction of execution concerns to package installation and therefore a coupling between the two:

1. On other platforms, it's very useful to be able to edit code within dependencies on the fly, either for experimentation or as temporary patches whilst waiting for an upstream release (I've had to do this before with Ruby gems). In order to support this, we'd need a way of invalidating the script cache before execution. I guess this could be as easy as just deleting the file and have scriptcs.exe detect this and recreate it at runtime. It's an extra step for the developer to have to remember but it's probably workable. I guess a hash of the dependency file timestamps could also be built at installation time and used to invalidate, but then you'd need an I/O hit to examine the dependency file timestamps at runtime.
1. Knowledge of which engine is going to be used for execution may be required at installation time. E.g. even if we restrict this feature to C#, I may choose to use my own C# engine with some custom features which are incompatible with the default method of script merging which means a) my engine would have to expose the relevant hooks for merging and b) I would have to specify the containing module at installation time.
1. Given that the merging may need to be engine specific, if engine implementations vary enough, I won't be able to install the package and then run it with engine A and then engine B unless they are compatible 'enough'. I would have to clobber the packages folder and reinstall, passing my engine B module as a parameter. I guess scriptcs.exe could also retain knowledge of which engine was used to build the cache and rebuild it when a different engine is used at runtime. I concede that this may be an edge case.
1. It makes it more difficult for someone to use an alternate package manager, e.g. (off the top of my head) a third party spm.exe based on Paket, since it would have to hook into the relevant engine at installation time to work out how to parse and merge the scripts and it would also require knowledge of the scriptcs module system and their potential installation locations (local/global). An interesting side note to this is that, currently, a third party package manager doesn't even have to be written in .NET. With the introduction of installation time script merging it would need to have the ability to load .NET assemblies to get at the engines within them. The added complexity and coupling may indeed prevent a third party package manager from ever being built but I'm not sure if this a concern or not. It comes down to how loosely coupled we want packaging and execution to be. E.g. in the node world, the coupling is very loose. So long as the file structure is right, that's all that npm has to worry about.

One more thing to consider is nested dependencies (https://github.com/scriptcs/scriptcs/issues/767). This can come later as a separate feature, but I think it's worth bearing in mind when thinking about our design so we can at least envisage a path towards it without too much rework.

I think this is an awesome feature and a huge step in the evolution of scriptcs as a platform, which is why I'm putting so much effort into the discussion. If I'm focusing on the wrong things then I'm relying on you all to put me straight! Sorry for the wall of text but it's as succinct as my limited writing skills allow :wink:.

glennblock:
WOW that is some reply! Glad you are excited about it. To be honest I was
ready to drop this feature based on the hurdles I hit with the initial
spike. It was only your prodding and saying you really wanted this that
pushed me to take a different approach!

Thanks for your attention and detail on the design. I am impressed you did
this without looking at hte code :-). You raised some really good
questions. I think a few of the main concerns can be addressed, though
having multi-language support out of the box may not be one. As I'll
mention several times below, I think that is OK :-).

I'll answer each question as best as I can. I hope you are ready for an
equally long reply :p

First to set context. My goal here was to make a good leap forward in
supporting a way to package scripts as packages and to do it in a
minimalist fashion, that is without breaking the entire design that exists
today. I also was very cognizant of minimizing the impact at runtime. It is
not intended to necessarily support every possibility in V1.  It does have
constraints and limitations. That's perfectly OK as it can be expanded over
time, and it is opt-in. Less is more in this case. if we start trying to
solve every problem, I have a big fear this will get bigger and bigger,
introduce a lot of friction into the code base, and will potentially not
get done period.

*"You mention that it is still a work in progress. What is left to be done?
I.e. if we were to merge now, would it be releasable?"*

It is not a work in progress. I put that comment when it started. This is
in my opinion releasable at least that was the goal.

*"You mention that it closes both #909
<https://github.com/scriptcs/scriptcs/issues/909> and #238
<https://github.com/scriptcs/scriptcs/issues/238>. It seems as though #909
<https://github.com/scriptcs/scriptcs/issues/909> supersedes #238
<https://github.com/scriptcs/scriptcs/issues/238> favouring 'scripts as
packages' (or whatever we end up calling them) over 'script packs as
script'. If so, I think we should close #238
<https://github.com/scriptcs/scriptcs/issues/238> as a wontfix."*

Or maybe 238 should just be the issue as that is actually what this is.
Good call though. I'd say 238 gets closed as fixed as well not 'wontfix'

*"I agree that we need to be careful with the naming. Perhaps, as part of
avoiding confusion with 'script packs', we should consider a different
naming convention for the entry level script than 'XXXPack.csx'? Perhaps
'XXXMain.csx'?*

I like "XXXMain.csx", I'll go with that since "Main" is familiar to C#
developers.

*"How does this work when another language engine is used at runtime? E.g.
if I run using the F# engine, are the content/scriptcs scripts simply
ignored due to their file extensions?"*

Yes you are correct. It will be ignored. I think that is OK for V1.

*"I'd also like to open up a conversation about the merged script caching
at installation time which, whilst ostensibly an implementation detail, I
think may have significant functional implications. It may indeed be the
right solution but I think it's important to consider the implications,
ensure we are comfortable with it and that we are doing it for the right
reasons. The more of @scriptcs/core
<https://github.com/orgs/scriptcs/teams/core> that can chime on this, the
better.*

*Firstly, is it proven that the caching is required? Is this purely a
performance optimisation? If so, how long does the script loading and
merging take? If it moved from installation time to runtime what would be
the performance hit? Other platforms do indeed read their dependencies from
the filesystem at runtime with acceptable performance so I'm interested to
know how scriptcs would fair with that approach. (Note that if merging
happened at runtime then writing the merged script to disk would serve no
purpose, apart from perhaps as a debugging aid when -debug is passed.)"*
First, I started off exploring doing this at runtime and ran into a bunch
of issues and concerns which this addresses. One big concern was
performance. Another was being minimally invasive. We do a lot of scanning
on startup, especially if there are a lot of packages, I did not want to
make that worse. I looked at introducing the discovery of scripts in the
initial scan, but it was going to break a lot of code and existing
contracts. Once you go this path, you cannot turn back with out a lot of
pain. I had an explicit goal to have something minimally invasive at the
start. With the current design, the impact at runtime is very small, as a
single script is loaded up with #load. It is a few lines of code in the
executor, and it only executes if the combined script is present. This
means there is no additional overhead at runtime if you do  not use this
feature, and if you do use the feature, well it just #loads a single script.

I think it is proven based on the amount of times we are hitting disk now.
As we're talking through this though, an interesting possibility is to move
the composing to runtime and only start composition IF the composed script
is not present. This might add additional work on the first execution, but
subsequent invocations would be faster. This offers another benefit that
since modules are loaded / it might open the door better for extensibility.
I do agree that it would be nice to NOT have to couple this to install.

I'll try moving it to runtime and see how this looks.


*"Knowledge of which engine is going to be used for execution may be
required at installation time. E.g. even if we restrict this feature to C#,
I may choose to use my own C# engine with some custom features which are
incompatible with the default method of script merging which means a) my
engine would have to expose the relevant hooks for merging and b) I would
have to specify the containing module at installation time."*

I don't consider this a mainstream case, and definitely not a goal for V1.
I would like long term for this feature to support multiple languages and
custom engines, but today it is restricted to C#. I think that is
acceptable for a V1 in order to get something out there which will work for
both Mono and .NET. So in summary, I'd say V1 supports C# with Roslyn or
the Mono engine.

The current design does have some support for different languages, but it
requires that both the main scripts and the package scripts are the same
language. Still there are gaps even to getting this to work now. If we move
the

Beyond this is the question of supporting multiple languages at the same
time for example: My script is C#, I have packages that are F# and say
IronPython. How can those work together? There's a boat load of issues in
making that work in particular if there are interdependencies. I don't at
all want to address this now as it opens up a huge can of worms. It may not
even be reasonable to ever get there.

That doesn't mean this work should not push forward.

*"Given that the merging may need to be engine specific, if engine
implementations vary enough, I won't be able to install the package and
then run it with engine A and then engine B unless they are compatible
'enough'. I would have to clobber the packages folder and reinstall,
passing my engine B module as a parameter. I guess scriptcs.exe could also
retain knowledge of which engine was used to build the cache and rebuild it
when a different engine is used at runtime. I concede that this may be an
edge case."*

As I mentioned above, I think it is acceptable to say for the initial
version that it only supports C#. 99% of scriptcs users use us for C#. I am
not disqualifying other languages, but I don't see it as a requirement
initially to have any progress.

*"It makes it more difficult for someone to use an alternate package
manager, e.g. (off the top of my head) a third party spm.exe based on
Paket, since it would have to hook into the relevant engine at installation
time to work out how to parse and merge the scripts and it would also
require knowledge of the scriptcs module system and their potential
installation locations (local/global). An interesting side note to this is
that, currently, a third party package manager doesn't even have to be
written in .NET. With the introduction of installation time script merging
it would need to have the ability to load .NET assemblies to get at the
engines within them. The added complexity and coupling may indeed prevent a
third party package manager from ever being built but I'm not sure if this
a concern or not. It comes down to how loosely coupled we want packaging
and execution to be. E.g. in the node world, the coupling is very loose. So
long as the file structure is right, that's all the at npm has to worry
about."*

Again, i would say that is fine for V1 to not have alternate package
manager support in scriptcs.exe. It should be possible however to wire it
up in a custom host. The work of combining the scripts is not done in the
engine and not dependent on NuGet. There is a new service called a
PackageScriptsComposer that does all the work. This API does not know
anything about NuGet, it only know about our package management
abstractions. Thus it should work fine with an alternate package manager as
long as it is implementing our abstractions.

Yes in the node world coupling is loose. But we also have -install in
scriptcs so we already broke that model. During install we create a
packages.config which instructs scriptcs what to load. We already have
coupling there.

I am not choosing to couple just because, the decision was based on
preserving performance and being minimally invasive. However, it may well
be that moving composing of the scripts to runtime and doing it
conditionally may be the answer.

*"One more thing to consider is nested dependencies (#767
<https://github.com/scriptcs/scriptcs/issues/767>). This can come later as
a separate feature, but I think it's worth bearing in mind when thinking
about our design so we can at least envisage a path towards it without too
much rework."*

Something to think about yes, something to address now? In my opinion no.
And though it sounds like it "could" work, it actually might not. It needs
quite a bit more investigation. I would not hold back this feature based on
whether it might support something we don't even understand yet. We'll
cross that bridge when we get there. I do see it has potential, as the
feature creates a wrapper for each package, and that could include
dependent things as well. There's a bunch of details to work out.

Here are my action items I am taking away:

1. I'll change the entry level csx to use the XXXMain convention.
2. I'll move the composition to runtime. This may actually allow supporting
different languages (I.e. F# script and F# package scripts) via a module.

Look for more commits shortly...

Thanks!
Glenn

glennblock:
OK @adamralph. I've moved composition of scripts to runtime rather than at load time. I've also renamed the entry script. Currently the composition only happens if the composed file is not present. That code is in a single place, so its not difficult to change it to load all the time. 

As part of this change, it should now be possible to replace the PackageScriptsComposer with a custom composer via module or in a custom host. The entry filename is now a virtual prop on the composer, thus an FSharpPackageScriptsComposer could change the entry name to say "XXXMain.fs".

glennblock:
Forgot to mention, the composition is invoked within ExecuteScriptCommand and ExecuteReplCommand. If one is a custom host, they will need to invoke PackageScriptsComposer explicitly to have it compose.

glennblock:
I've updated the design doc with the new approach: https://github.com/scriptcs/scriptcs/wiki/Package-scripts

adamralph:
Thanks for the detailed reply @glennblock! All the points you make are reasonable and thanks for making the additional changes. I'll start digging into the implementation soon...

One thing I noticed from the wiki page:

> Note: var cannot be used here for the logger this is code that will be sitting in a wrapper class so it cannot be loose code. The above code sets a class level private field logger. Require<T> in this context is a static method on the wrapper class. See the section "Scoping / Containment" below for more on the wrapper.

I understand the technical limitation, but it would be great if we could find a way to avoid this. It makes the code quite unnatural and I wonder if there are other implications on authoring which we haven't thought of. Ultimately, the author is filling in the body of a class without the explicit class structure written around it. I'm just wondering if it may be a strange experience to get used to for authors. Personally (and yes I am fussy :wink:) I'd find it a bit of an eyesore to have to use explicit types everywhere instead of `var`.

Just to throw it out there (not fully thought through) would it be possible for us to make the class definition explicit? I.e. the author simply writes a regular class in the entry level script? I guess `Require` or anything else which we'd otherwise expose on the wrapper class would be a challenge. Perhaps that could be a static somewhere? We could force inheritance from a base but that's quite ugly. Another option would be constructor injection but then we'd have to provide a factory for scripts to create an instance, i.e. something similar to `Require`, or perhaps `Require` could even serve both purposes. Mind you, I like the fact that currently you can just `new` the class so it would be a shame to lose that. Anyway, just throwing some ideas out there. Interested to know if anyone else can come up with something here.

adamralph:
Note: we have some beta testing volunteers taking the new feature for a test drive so we should attempt get some feedback from them before merging.

glennblock:
@adamralph Having the wrapper establishes a convention and forces isolation/scoping. Also it ensures that the proper base functionality is there as the wrapper derives from a base class for providing things like Require<T>.

I agree it would be nice for "var" to work and there might be some Roslyn magic to make that possible. However I think it will not be trivial, especially to also support mono. Beyond that I don't think its necessary now.

glennblock:
I made "-Install" now delete the Script Libaries file so that you can force it to be recomposed.

glennblock:
@adamralph I think we should merge this. If usability feedback comes in, I will do a follow up PR. There is no need imo to hold this back. It's opt-in so it will only affect those that decided to install Script Libraries.

src/ScriptCs.Core/ScriptExecutor.cs
```diff
@@ -10,8 +12,11 @@ namespace ScriptCs
 {
     public class ScriptExecutor : IScriptExecutor
     {
-        public static readonly string[] DefaultReferences = new[] { "System", "System.Core", "System.Data", "System.Data.DataSetExtensions", "System.Xml", "System.Xml.Linq", "System.Net.Http" };
+        public static readonly string[] DefaultReferences = new[] { "System", "System.Core", "System.Data", "System.Data.DataSetExtensions", "System.Xml", "System.Xml.Linq", "System.Net.Http", typeof(ScriptExecutor).Assembly.Location };
```
filipw: is this necessary? why do scripts need to know about ScriptCs.Core?

src/ScriptCs.Core/ScriptLibraryComposer.cs
```diff
@@ -0,0 +1,117 @@
...
+            _packageContainer = packageContainer;
+            _packageAssemblyResolver = packageAssemblyResolver;
+            _logger = logger;
+        }
+
+        internal string GetMainScript(IPackageObject package)
+        {
+            var script =
+                package.GetContentFiles()
+                    .SingleOrDefault(f => f.EndsWith("main.csx", StringComparison.InvariantCultureIgnoreCase));
```
filipw: what happens if i have two "mains"?

src/ScriptCs.Core/ScriptLibraryComposer.cs
```diff
@@ -0,0 +1,117 @@
...
+            get { return "ScriptLibraries.csx"; }
+        }
+
+        public void Compose(StringBuilder builder = null)
+        {
+            var namespaces = new List<string>();
+            var references = new List<string>();
+
+            var packagesPath = Path.Combine(_fileSystem.CurrentDirectory, _fileSystem.PackagesFolder);
+            var packageReferences = _packageAssemblyResolver.GetPackages(_fileSystem.CurrentDirectory);
```
filipw: we shoud probably cache this in the package assembly resolver, I think we are asking for this a few times

src/ScriptCs.Core/ScriptLibraryComposer.cs
```diff
@@ -0,0 +1,117 @@
...
+        {
+            _logger.DebugFormat("Finding package:{0}", reference.PackageId);
+            var package = _packageContainer.FindPackage(packagesPath, reference);
+            _logger.Debug("Package found");
+            var script = GetMainScript(package);
+            if (script != null)
+            {
+                script = Path.Combine(packagesPath, string.Format("{0}.{1}", package.Id, package.TextVersion), script);
+                _logger.Debug("Pre-processing script");
+                var result = _preProcessor.ProcessFile(script);
```
filipw: the code here does not handle the `LoadedScripts` property of `FilePreProcessorResult` - is this by design?

src/ScriptCs.Core/ScriptLibraryWrapper.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using ScriptCs.Contracts;
+
+namespace ScriptCs
+{
+    public abstract class ScriptLibraryWrapper
```
filipw: this class does not surface `Env` - I know it's not on `IScriptHost` but it might be a good moment to ask if it should be?

src/ScriptCs.Core/ScriptExecutor.cs
glennblock: Because the ScriptLibraryWrapper is in ScriptCs.Core and that is what the wrapper class that is generated derives from. We already added this for Roslyn and were not doing it for Mono. Now it has just been upleveled.

src/ScriptCs.Core/ScriptLibraryComposer.cs
```diff
@@ -0,0 +1,117 @@
...
+        public void Compose(StringBuilder builder = null)
+        {
+            var namespaces = new List<string>();
+            var references = new List<string>();
+
+            var packagesPath = Path.Combine(_fileSystem.CurrentDirectory, _fileSystem.PackagesFolder);
+            var packageReferences = _packageAssemblyResolver.GetPackages(_fileSystem.CurrentDirectory);
+            var packageScriptsPath = Path.Combine(packagesPath, ScriptLibrariesFile);
+
+            if (!_fileSystem.DirectoryExists(packagesPath) || _fileSystem.FileExists(packageScriptsPath))
```
filipw: what happens if I:
1) install script lib from nuget, run scriptcs
2) install another script lib afterwards
It seems like it will not be picked up because a packageScriptsPath will already exist?

src/ScriptCs.Core/ScriptLibraryComposer.cs
glennblock: Currently it will just end up ignored. It probably makes sense though to add warning. Another option is allowing 2 mains, but I don't currently think that makes a lot of sense

src/ScriptCs.Core/ScriptLibraryComposer.cs
glennblock: Yes, I've had thoughts about caching this, but I don't think we need to do that now as this only runs when the script is composed as after it exists it is not rebuilt unless one runs `scriptcs -install`

src/ScriptCs.Core/ScriptLibraryComposer.cs
glennblock: What needs to get handled? I am not following.

filipw: CSX could have a #load which `FilePreProcessorResult` would surface  via `LoadedScripts` but it's ignored here

src/ScriptCs.Core/ScriptLibraryWrapper.cs
glennblock: In my opinion it should be there, though it will be a breaking change if we add it to the interface. It will only implement those who have a custom host though.

Today custom scriptcs like "ScriptCs.Logger" do rely on this to get the args, so maybe we should add it.

src/ScriptCs.Core/ScriptLibraryComposer.cs
glennblock: -install clears the packageScriptsPath. I added this logic.

glennblock: See here: https://github.com/scriptcs/scriptcs/pull/928/files#diff-1f6b4857c2b5602232a7abf410ba4675R48

src/ScriptCs.Core/ScriptLibraryComposer.cs
glennblock: In the acceptance tests CalculatorMain.cs has a #load of Multiply.csx and that works fine. Where is the gap? Why do I care about LoadedScripts, is this to set the env or something?

src/ScriptCs.Core/ScriptLibraryComposer.cs
glennblock: I'll add the warning logic. I think it makes sense to think of the wrapper as a namespace, thus having only one makes sense. You can always have public classes inside the wrapper which can be created if you want to have multiple.

glennblock:
I'll add "Env" to "IScriptHost" and also add logic to log a meaningful error if a Script Library has 2 mains.

paulbouwer:
I'm currently testing some functional scenarios. I think between @adamralph and @filipw you should have the code and architecture side covered.

@glennblock - I'm using your `CalcTest.csx` sample. It is deployed to `C:\Dev` on my machine.

I get the following error when I'm not in the `C:\Dev` folder but calling from another folder and providing a path to the script.

```
C:\ > scriptcs C:\Dev\CalcTest.csx
ERROR: Script compilation failed: C:\Dev\CalcTest.csx(1,16): error CS0246: The type or namespace name 'Calculator' could not be found (are you missing a using directive or an assembly reference?).
```

It works if called from within the script folder.

```
C:\Dev > scriptcs CalcTest.csx
7
```

This behaviour is consistent when used with the `-Script` parameter

```
C:\ > scriptcs -Script C:\Dev\CalcTest.csx
ERROR: Script compilation failed: C:\Dev\CalcTest.csx(1,16): error CS0246: The type or namespace name 'Calculator' could not be found (are you missing a using directive or an assembly reference?).
```

src/ScriptCs/ScriptCsArgs.cs
```diff
@@ -56,6 +56,10 @@ public ScriptCsArgs()
         [ArgDescription("Allows installation of packages' prelease versions")]
         public bool AllowPreRelease { get; set; }
 
+        [ArgShortcut("rsl")]
+        [ArgDescription("Refresh the cache of script libraries")]
+        public bool RefreshScriptLibraries { get; set; }
```
paulbouwer: There seems to not be any code wired up to this option.

src/ScriptCs.Core/ScriptLibraryComposer.cs
paulbouwer: I modified `CalculatorMain.csx` by adding `..` to the logged message as follows:

```
public double Add(double a, double b) {
  logger.Info(String.Format("Adding .. {0} + {1}", a, b));  <==
  return a+b;
}
```

Running it results in:

```
c:\Dev > scriptcs CalcTest.csx -- -loglevel info
[2015-02-28T11:39:35]  INFO  Adding .. 5 + 2
7
```

I deleted the `ScriptCs.Calculator.0.1.0` folder in the `scriptcs_packages` folder to force a re-install of the package.

```
c:\Dev > scriptcs -install
WARN: Cannot find: ScriptCs.Calculator 0.1.0.0
Installing packages...
Installed: ScriptCs.Calculator 0.1.0.0
Package installation succeeded.
Saving packages in scriptcs_packages.config...
Updating scriptcs_packages.config...
Skipped ScriptCs.Calculator because it already exists.
Skipped ScriptCs.Contracts because it already exists.
Skipped ScriptCs.Logger.ScriptPack because it already exists.
Successfully updated scriptcs_packages.config.

c:\Dev > scriptcs CalcTest.csx -- -loglevel info
[2015-02-28T11:40:11]  INFO  Adding .. 5 + 2
7
```

The function was not replaced with the original indicating that the `ScriptLibraries.csx` had not been deleted. @glennblock - If you check the line referenced in your note above you'll notice that the delete is being logged but not actioned.

```
if (_fileSystem.DirectoryExists(packagesFolder))
{
    _logger.DebugFormat("Deleting: {0}", scriptLibrariesFile);
}
```

It also seems as though the `RefreshScriptLibraries` option is not wired up. I can't find any code that references `RefreshScriptLibraries` from `src/ScriptCs/ScriptCsArgs.cs`.

Running the following does not refresh the cached Script Library.

```
C:\Dev > scriptcs CalcTest.csx -RefreshScriptLibraries -- -loglevel info
[2015-02-28T11:58:08]  INFO  Adding .. 5 + 2
7
```

glennblock: Good catch on the delete. The refresh option should not be there at all. It
was supposed to be pulled in favor of -install having the logic.

paulbouwer: I think that is is a better option. Refresh it all as part of any additional changes to the packages. And remove the `RefreshScriptLibraries` option which may be confusing.

Will you be able to determine if the package being installed is a Script Library and only refresh then ? Or do you think the extra work involved there is worth it ...?

paulbouwer:
I've just finishing testing on Linux too. And all looks good. The hardest part was getting NuGet and multiple repositories working properly on Linux.

src/ScriptCs.Core/ScriptLibraryComposer.cs
glennblock: No, it will just delete the file if you do install. I think I know what happened, I forgot to remove refresh from the args class, I will do that now.

src/ScriptCs.Core/ScriptLibraryWrapper.cs
glennblock: I am adding it to IScriptHost. This means moving ScriptEnvironment to Contracts, OR having an IScriptEnvironment in Contracts. As ScriptEnvironment is so light, I am thinking to just move it to contracts.

src/ScriptCs/ScriptCsArgs.cs
glennblock: Yes this was left accidentally as I pulled the new command but forgot to change the args. Fixing now.

src/ScriptCs.Core/ScriptLibraryWrapper.cs
glennblock: One caveat here is this requires the MonoHost to have an explicit implementation of Env as it already has a public static property. I added this and it compiles fine.

glennblock:
@paulbouwer @filipw thanks for the review! I applied both your feedback with the last PR.

* Clear warning generated / Script Library is ignored if there are 2 Main files
* `-Install` actually does delete `ScriptLibraries.csx` :-)
* Added `Env` to `IScriptHost` implementing new `IScriptEnvironment` interface.
* Added `Env` to `ScriptLibraryWrapper` so it surfaces to Script Library authors.

glennblock:
@paulbouwer it looks like the issue is that it is looks relative to the CurrentDirectory from where scriptcs is run to load the Script LIbrary, rather than relative to where the script is located. I'll need to fix this. Thanks for finding this!

glennblock:
OK realized another missing piece the REPL was not wired up.

glennblock:
@paulbouwer this should be fixed now. It will work if you are passing a non-rooted path. I have added an acceptance test which verifies this. Also REPL support is now in also with an acceptance test.

paulbouwer:
@glennblock - retested on Windows and Linux using your latest commits:

- 865360623002ca873cb4dd922936d27f17acc596
- cfddb58ea659796b8c1d4a3a5cf780b2c93a5935

Issues I raised are resolved.

glennblock:
@paulbouwer great!

glennblock:
@filipw what's left? I don't think we need to address the caching now as all the Script Libraries are cached when the ScriptLibraries.csx file is persisted. The only hit will happen when that file is either built the first time, or after a `scriptcs -install` which I think is acceptable for now.

test/ScriptCs.Tests.Acceptance/ScriptLibraries.cs
```diff
@@ -0,0 +1,104 @@
...
+
+            "Then I see INFO outputted from the required Logger script pack"
+                .f(() => output.ShouldContain("INFO"));
+        }
+
+        [Scenario]
+        public static void UsingAMethodInAScriptLibraryInTheRepl(ScenarioDirectory directory, string output)
+        {
+            var scenario = MethodBase.GetCurrentMethod().GetFullName();
+            var args = new[] {"r"};
```
adamralph: This should be `-r`.

test/ScriptCs.Tests.Acceptance/ScriptLibraries.cs
```diff
@@ -0,0 +1,104 @@
...
+
+            "Given a script which uses ScriptCs.Calculator"
+                .f(() => directory = ScenarioDirectory.Create(scenario)
+                    .WriteLine("foo.csx", @"Console.WriteLine(""Type:"" + new Calculator().GetType().Name);" + Environment.NewLine + "Environment.Exit(0);"));
+
+            "And ScriptCs.Calculator is installed"
+                .f(() => ScriptCsExe.Install("ScriptCs.Calculator", directory));
+
+            "When executing the script"
+                .f(() => output = ScriptCsExe.Run("foo.csx", false, Enumerable.Empty<string>(), args, directory));
```
adamralph: The args and scriptArgs are in the wrong order here.

adamralph:
LGTM. @filipw do you still have any concerns?

I have just one question, after installing a script library, will it will no longer be possible to run a .fsx file from the same folder using the F# engine (or any other language) since the C# generated from the script libraries will be appended to the script regardless of which engine and language are being used? If so, I think it's OK to have this as a stated limitation for now, but I wonder if we can see a path out of this? Would we have to move the `ScriptLibraryComposer` into the engines?

src/ScriptCs.Core/ScriptLibraryComposer.cs
```diff
@@ -0,0 +1,129 @@
...
+                builder.Insert(0, String.Format("#r {0}{1}", reference, Environment.NewLine));
+            }
+            _fileSystem.WriteToFile(packageScriptsPath, builder.ToString());
+        }
+
+        protected internal virtual void ProcessPackage(string packagesPath, IPackageReference reference, StringBuilder builder, List<string> references,
+            List<string> namespaces)
+        {
+            _logger.DebugFormat("Finding package:{0}", reference.PackageId);
+            var package = _packageContainer.FindPackage(packagesPath, reference);
```
adamralph: If a package is missing from the packages directory, a null reference is returned here. It is then passed `GetMainScript()`, which throws an NRE, which kills the app.

adamralph:
Found this just now:

image

Because the type of `ScriptHost.Env` has changed from `ScriptCs.ScriptEnvironment` to `ScriptCs.Contracts.IScriptEnvironment`, we now need to add a reference to `ScriptCs.Contracts` to the script.

src/ScriptCs.Core/ScriptLibraryComposer.cs
glennblock: OK. I'll log a warning.

glennblock:
@adamralph Yes we do, otherwise there will be another fiasco like we had in the past with script packs breaking. Great catch!

test/ScriptCs.Tests.Acceptance/ScriptLibraries.cs
glennblock: OK, will fix.

test/ScriptCs.Tests.Acceptance/ScriptLibraries.cs
glennblock: Will fix!

glennblock:
@adamralph on F#:

A path that is possible now, is having F# script + F# script libraries, but it will require the F# module to register a custom composer.

As to multi-lang / mix and match, I am not sure but I have thought about it a little. One idea is we have a script per language that gets compiled. Each would have to get compiled and persisted so that they could be loaded up. It could be as simple as CsharpScriptLibraries.dll, FSharpScriptLibraries.dll.

Then there would have to be a smart loader that knows how to load the assemblies since they have different entry points and ways to instantiate. There's a bunch of cost concerns here to be worried about from a startup perspective. I think it is non-trivial, though something to think about for the future.

An alternative option that could work is to have multiple language files in the ScriptLibrary package and then load up the correct versions to merge in based on the engine / main script. So you could have .csx and .fs files for example.

This can totally work, but it pushes more work on the authoring side.

Let's not forget we still have binary script packs, so that is also another option.

glennblock:
@adamralph I'll get all the fixes in later today, thanks everyone for the time on reviewing this important feature!

glennblock:
@adamralph the last commit should fix your issue.

glennblock:
@adamralph I mean the issue with "Env".

glennblock:
With the last commit I am done with any outstanding issues that I am aware of.

test/ScriptCs.Tests.Acceptance/ScriptLibraries.cs
glennblock: fixed

src/ScriptCs.Core/ScriptLibraryComposer.cs
glennblock: fixed

adamralph:
@glennblock thanks for the latest commits. I've taken a look through them and left some questions above. The logging enquiry is minor, and the static vs non-static is purely educational. The only important one is about the special casing in the assembly resolution, which I'd like to get an understanding of.

glennblock:
The assembly filtering is necessary because script packs include earlier versions of Contracts as a reference. If both are passed as references to the engine it will be blow, there can only be one. This fix ensures that there is one reference and it always the version that the runtime itself uses.

If this logic is not present you will get an error from the compiler as there are multiple versions referenced. I tested this exahaustively. 

The static on env is to allow access to the args when initializing a field.

glennblock:
On the logging, I'll add the package name, I went back and forth on it, but you are right, I should add it explicitly.

glennblock:
@adamralph I added PackageId to the log messages now.

adamralph:
Thanks, log message looks good. Static Env - of course! Good thinking.

adamralph:
@glennblock thanks for the explanation regarding the assembly filtering, that makes sense.

My concern with this is that it introduces an inappropriate responsibility to the `AssemblyResolver`,by special casing an assembly just because of something the `ScriptExecutor` is doing internally with its default references and if those ever change, `AssemblyResolver` also has to change. `AssemblyResolver` is also used elsewhere in the code base so I think it should be kept generalised.

I think this behaviour would be better encapsulated inside `ScriptExecutor` since it knows what its default references are and it also has public `AddReferences()` overloads which may be called from anywhere (including custom hosts). To that end, I've prepared a commit which does that https://github.com/adamralph/scriptcs/commit/c719bf55b07a6511bc89879c191ad3a2dce188a0. I've tested and it does also fix the problem. (Incidentally, if someone were ever to, accidentally or otherwise, install ScriptCs.Core in their packages folder, the commit also makes sure that won't break things since it also ensure only the default ScriptCs.Core reference is added.)

Can you take a look and see what you think? If you agree with the approach you can fast forward your branch to that commit to add it to the PR. In that case, I'd be happy to merge.

glennblock:
@adamralph I like it, this is a definite improvement. Can you send me a PR to my branch?

adamralph:
Assuming your branch hasn't moved on in the meantime, you can just do

```bash
git remote add adam https://github.com/adamralph/scriptcs.git
git checkout 909
git pull adam 909
git push origin 909
```

glennblock:
Doing this now. I just pulled locally and all tests built, just need to try a few things manually. If they work, I'll push. After which I am assuming we can merge? ;-)

adamralph:
I think so! :smiley:

glennblock:
OK, I beat it up. Works great, here it comes!

glennblock:
OK @adamralph, tag you are it.

glennblock:
Oh wait....your commit is not there hmm. I did the FF.

glennblock:
@adamralph nvm, your commit is there, I see it.

adamralph:
Travis CI acting the goat again... I'll try and re-run (https://github.com/scriptcs/scriptcs/pull/950 solves that btw).

glennblock:
@adamralph fancy docker stuff +1

glennblock:
Don't merge yet, need to update the test "text"

glennblock:
Thank God!

adamralph:
Great work @glennblock - can't wait to show this feature off

glennblock:
:100:

Issue #898, opened 2015-01-09
Fix inconsistencies in IFileSystem


adamralph:
Currently has inconsistent method names and parameter names:

* Some methods which have, or can potentially have, file and directory counterparts are missing `File` from the name of the file method, e.g. `Copy`, `Move`
* Most methods have `path` as the param name but a few have 'file' or 'filePath', when the name of the method already, or should already, indicate file or directory (see previous point)
* In a few cases `directory` has been abbreviated to `dir`, which is inconsistent
* `FileDelete` is {NounVerb} which is inconsistent with all other methods which are {VerbNoun}, it should be `DeleteFile`
* `WriteAllBytes` should be renamed as an overload of `WriteToFile`

glennblock:
A bunch of folks have custom file system implementations. They will all break with this change. I know its messy, but do we really _need_ to do it.

adamralph:
For the long term, I believe it's important to clean up the API. I agree that we have other work which should take priority in the short term, so we can mark it low priority.

The approach I would take is to add the new methods, keep the legacy methods, deprecate them by adding `[Obsolete]`, and then remove them on the next major version bump.

glennblock:
@adamralph that doesn't work well for interfaces in that it would still require both to be implemented. 

For V1 cleanup makes sense to consider.

adamralph:
Ah yes, that's very true for interfaces.

srutis90:
I'll take it.

PR #933, opened 2015-02-16
#898 fix naming inconsistencies in IFileSystem methods and parameters

srutis90:
The parameter renames are just meant to be consistent with the rest of the file - does not break anything.
The method renaming are breaking changes. Marking them obsolete on the interface would still need the custom implementations implement both old and new API methods. 
Hence, did a rename, so that when the others pull, they also just rename the methods in their custom FileSystem implementations.

### DIFF
--- a/src/ScriptCs.Contracts/IFileSystem.cs
+++ b/src/ScriptCs.Contracts/IFileSystem.cs
@@ -7,13 +7,13 @@ namespace ScriptCs.Contracts
     public interface IFileSystem
     {
         IEnumerable<string> EnumerateFiles(
-            string dir, string search, SearchOption searchOption = SearchOption.AllDirectories);
+            string path, string searchPattern, SearchOption searchOption = SearchOption.AllDirectories);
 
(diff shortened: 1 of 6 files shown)


glennblock:
@srutis90 thanks for putting in the effort, but I don't think we want to do this now. As I mentioned in the original issue, this introduces a breaking change. We might decide to break it for 1.0, but not right now.

glennblock:
@srutis90 please wait for confirmation from a member of the core team on a "You Take It" item in the future before before doing the work.

Thanks
Glenn

adamralph:
#898

srutis90:
@glennblock  Thank you for that pointer on waiting for a reply ! 

Sruti

glennblock:
Hi @Sruti

I realize we need to clarify this better in our docs which I am going to
fix.

The hot button on this PR is the breaking change. We're very sensitive to
this as we've had breaking changes in the past that broke a lot of people
using us and it was painful.

Thanks for your patience and effort!

PR #944, opened 2015-03-02
Rewrite of Program.Main, CLI args handling and config file reading #897

### LINKED ISSUE #897: CLI args/config file refactoring
IMO this whole space is currently quite confusing and clunky. After spending time on https://github.com/scriptcs/scriptcs/pull/896 I started to think about an alternative approach. Here's what I've come up with so far:

1. Splitting of the CLI args into command args and script args is lifted up to `Program.Main`. I think this can be re-written trivially so it won't be necessary to encapsulate it in a type. It's really a `Program.Main` level operation.
1. Parsing of the command args into `ScriptCsArgs` is lifted up to `Program.Main`. Normally this is a simple call to `PowerArgs` but since this also has to do some gymnastics regarding `-install` which has two optional parameters, package name and version (something which I think is not natively supported by PowerArgs), this is probably best encapsulated in a `ScriptCsArgsFactory`. If `ScriptCsArgsFactory` throws an exception, `Program.Main` writes the exception message to the console, shows usage and then exits with code 1 (as currently implemented in #896)
1. If `ScriptCsArgs.Version` or `ScriptCsArgs.Help` are `true`, `Program.Main` either shows the version or displays usage respectively and exits with code 0. This removes the need for `VersionCommand` and `ShowUsageCommand` which, IMO, are not commands, they are just options to show metadata without actually performing a command. This fits the behaviour of other platforms, e.g. node. For more on that, see https://github.com/scriptcs/scriptcs/issues/175#issuecomment-68596753
1. We read and store the `Config` property from `ScriptCsArgs`
1. We introduce a new type `ScriptCsConfig` and construct an instance using `ScriptCsArgs`. This can be encapsulated in a `ScriptCsConfigFactory` if required

Once these steps have been completed, the concept of 'arguments' no longer exists (disregarding the script args which need to be passed all the way through). We only have to think about rehydrating `ScriptCsConfig` instances from the local and global config files and then performing a masking operation: commandLineScriptCsConfig + localScriptCsConfig + globalScriptCsConfig. All properties on `ScriptCsConfig` should be nullable so for each property we take the first non-null value, falling back to a default.

We then pass the final `ScriptCsConfig` instance, along with the script args, to `CommandFactory` and the rest of the code remains as is.

### LINKED ISSUE #945: Fix profile optimization
Profile optimization is having no effect because the call to `System.Runtime.ProfileOptimization.SetProfileRoot` is being with the location of the executable as the `directoryName`. It should in fact be the name of the directory where the executable is located.

### LINKED ISSUE #962: Remove extra line breaks from scriptcs -help
For some reason, there is an extra line break between each line.

```batch
C:\code\scripts>scriptcs -help
Usage: scriptcs options

   OPTION                    DESCRIPTION

   -ScriptName (-script)     Script file name, must be specified first

   -Repl (-R)                Launch REPL mode when running script. To just launch REPL, simply use 'scriptcs' without any args.

   -Help (-?)                Displays help

   -Debug (-D)               Emits PDB symbols allowing for attaching a Visual Studio debugger [default=False]

   -Cache (-C)               Flag which determines whether to run in memory or from a .dll [default=False]

   -LogLevel (-log)          Flag which defines the log level used.

   -Install (-I)             Installs and restores packages which are specified in packages.config

   -Global (-g)              Installs and restores global packages which are specified in packages.config

   -Save (-S)                Creates a packages.config file based on the packages directory

   -Clean (-Cl)              Cleans installed packages from working directory

   -AllowPreRelease (-pre)   Allows installation of packages' prelease versions

   -Version (-V)             Outputs version information

   -Watch (-W)               Watch the script file and reload it when changed

   -Modules (-M)             Specify modules to load

   -Config (-Co)             Defines config file name [default=scriptcs.opts]

   -PackageVersion (-P)      Defines the version of the package to install. Used in conjunction with -install

   -Output (-O)              Write all console output to the specified file


   EXAMPLE: scriptcs server.csx -logLevel debug
   Shows how to run the script and display detailed log messages. Useful for debugging.



C:\code\scripts>
```

### LINKED ISSUE #963: Speed up -help and -version
These should be near instantaneous, all they have to is write the usage and version to the console.

### LINKED ISSUE #990: scriptcs -debug without -repl does not configure REPL correctly.
This is not new in 0.14. It is pre-existing in 0.13.3.

Leads to all kinds of problems, e.g.

```batch
mkdir bug-repro
cd bug-repro
scriptcs -install Nancy
scriptcs -debug
using Nancy;
```

> Index was outside the bounds of the array.

Without `-debug`, there is no error.

adamralph:
fixes #990
fixes #897
fixes #945
fixes #962
fixes #963


glennblock:
Please wait on doing any big merges until the Script LIbrary PR goes in.

adamralph:
Incidentally, `-version` and `-help` now execute instantly (previously each had a 2 second delay on my machine) since they are effectively a no-op with regard to the application.

Help output is also much nicer, since the extra line breaks have gone away:

image

adamralph:
I rebased this after the script libraries changes went into dev (there were no conflicts)

adamralph:
This is good to go now.

adamralph:
rebased, FWIW

adamralph:
@scriptcs/core it would be great if someone could have a look at this today. I'm flying to Portugal tomorrow afternoon and I'll be largely dark for a week.

It's a big change set, but each commit is very focused so reviewing one commit at a time is probably going to be easiest and quickest.

I'd love to see this go in. It's a massive simplification of the entry point to the app.

filipw:
does it handle stuf like `scriptcs -install foo 4.0.5` ?

src/ScriptCs/Program.cs
```diff
@@ -12,41 +10,51 @@ internal static class Program
...
             }
 
-            var scriptServicesBuilder = ScriptServicesBuilderFactory.Create(arguments.CommandArguments, arguments.ScriptArguments);
-            var factory = new CommandFactory(scriptServicesBuilder);
-            var command = factory.CreateCommand(arguments.CommandArguments, arguments.ScriptArguments);
-            return (int)command.Execute();
-        }
+            if (commandArgs.Help)
+            {
+                Console.WriteLine(ScriptCsArgs.GetUsage());
```
filipw: how is it possible it now shows up without new lines?

filipw:
does it handle starting REPL with log level set?

filipw:
and a non-educated comment - you have deleted lots of tests; I understand the corresponding classes are gone, but do we no longer need to check the scenarios they covered?

adamralph:
Guys, thanks very much for reviewing! It was nice to come home this evening to all these comments. 

In response:

@filipw 

> does it handle stuf like scriptcs -install foo 4.0.5 ?

Yes, that is handled inside `ScriptCsArgs.Parse()` and is unit tested.

> does it handle starting REPL with log level set?

Yes, this has been handled in `CommandFactory` since https://github.com/scriptcs/scriptcs/issues/869 went out in 0.13. That code has not been changed.

> and a non-educated comment - you have deleted lots of tests; I understand the corresponding classes are gone, but do we no longer need to check the scenarios they covered?

There is way less code now and it is much simpler. The work now takes place in three places:-

* `ScriptCsArgs.Parse()` - the only meaningful thing that happens inside here beside wrapping around `PowerArgs.Args.Parse<T>()` is the special casing for `-install` to allow it to take a package name and a version. This is unit tested in `ScriptCsArgsTests`.
* `ConfigMask` a trivial type with a static factory method and a `File.Exists` wrapper around a JSON.NET deserialization call
* `Config.Apply()` applies a `ConfigMask` to a `Config` and returns a new `Config`

I think all but one of the scenarios which these types affect are covered in the acceptance tests. The only one which isn't, is the masking of CLI args over local and global options files and, come to think of it, that acceptance test should probably be added. My general approach to testing is to ensure the important scenarios are covered by acceptance tests and only to drill down into 'unit tests' when it's felt some particular implementation detail is risky enough to require it. That's why included a unit test for `ScriptCsArgs.Parse()`. This is really the only complicated bit of logic that has survived from the original implementation, the rest is rather trivial.

@glennblock

> This block of code that parses args should be in its own function and called by main. Makes the code easier to maintain.

Personally I would leave it where it is since it's trivial and single use. I only tend to encapsulate something if it's complicated or reused. If feelings are strong on this, I can encapsulate in a function.

src/ScriptCs/Program.cs
adamralph: I've really no idea. It's a nice side effect of the PR though :stuck_out_tongue_winking_eye:

glennblock: Adam I meant in a function not a sep class :-)

adamralph:
@glennblock sure, I meant function too, 'encapsulate' was the wrong word :smile: 

So to paraphrase what `Main` is doing:

* divide the args into those for me and those for the script
* read the args for me (if I can't understand them, show what I do understand and return an error code)
* if you want help, show it and return a success code
* if you want to know my version, show it and return a success code
* if you want to load a config file which doesn't exist, refuse and return an error code
* create a config based on what I've been told
* load configs from the global and local files and mask the local config over the global config and my config over that
* create a command factory, feed the config into it and execute any resulting commands

Which part(s) do you think should move to a function?

glennblock:
Lines 18 to 35, everything related to parsing the args

adamralph:
Parsing the args is already encapsulated in `ScriptCsArgs.Parse()` (second point above).

Tell you what, if this is the only thing you'd like to change, perhaps you can merge this PR and send your proposed refactoring in a new PR?

adamralph:
@filipw I've added Configuration scenarios to the acceptance tests

glennblock:
I am not finished reviewing. As to the refactoring it is reducing the code in Main and making it more readable. Extract Method is a very common pattern. The lines I mentioned can be easily ripped to a parse method.

adamralph:
Sure, I'm familiar with Extract Method :wink:.

I guess it's a matter of taste. IMHO the lines should stay where they are. The key aspect is the return codes. There are a few cases in the method which `Main` needs to check and return a 0 or 1 appropriately. I wouldn't move these elsewhere as I think that will make the code less readable, since you'd have to propagate the return codes back up the stack.

There is actually only one line which does arg parsing, and that is line 22. The heavy lifting there is already extracted into `ScriptCsArgs.Parse()`.

The only lines which I think could be moved elsewhere are:

* 16 and 17, but this would come with the overhead of either creating a return type, returning a `Tuple<T1, T2>` or using `out` parameters. All of these options I think would bloat the code for little benefit
* lines 49 to 52 - this is the loading and masking of the local and global config. Although it's 4 lines, it's a single statement with method chaining and I think it's very easy to read so I'd probably leave as is.

adamralph:
Ultimately `Main` is only 46 lines long and, as the entry point, it needs to orchestrate according to the CLI args with the constraint of returning status codes. In that context, I think it's as decomposed as is practical.

adamralph:
hold until post 0.14

adamralph:
0.14 has been released, so this PR is good to go again.

adamralph:
hold off - see https://github.com/scriptcs/scriptcs/issues/990

adamralph:
@glennblock `Program.Main` is now smaller :wink:

adamralph:
This is good to go again.

filipw:
:shipit:

glennblock:
:+1: 

Great work, much more simplified!

PR #954, opened 2015-03-05
implemented PackagesWithDuplicateAssemblies scenario - fix for #953

### LINKED ISSUE #953: Duplicate references cause script execution to fail
image

This is in the *current release*, i.e. 0.13.3 on Choco.

IIRC, we had some kind of 'de-duplication' of references at one time but it seems to have disappeared. I think what we require is a more general version of the changes that went into the dev branch recently for the scriptcs assemblies (https://github.com/scriptcs/scriptcs/blob/af885ff75cc31736bc60b298ce26c960d70b79b4/src/ScriptCs.Core/ScriptExecutor.cs#L66-L67 and https://github.com/scriptcs/scriptcs/blob/af885ff75cc31736bc60b298ce26c960d70b79b4/src/ScriptCs.Core/ScriptExecutor.cs#L88-L96) which take into account *all* the assemblies currently referenced.

Unless we can push out 0.14 soon I think this should be a patch on master (0.13.4) but I'd prefer someone else to reproduce and confirm the issue first.

adamralph:
fixes #953

In summary, all the duplicate reference logic, spanning both assemblies and paths is now encapsulated in `AssemblyReferences` which is an immutable data structure. All other objects no longer have this concern.

By far the quickest and easiest way to review this PR will be one commit at a time. Each is isolated with a specific focus.


khellang:
Looks good! :+1:

PR #965, opened 2015-03-07
Do not use initialize modules MEF catalog if the only module in play is Roslyn or Mono

### LINKED ISSUE #964: Only scan for modules when -m is specified
Currently we scan even when -m is not specified, since our automatic engine selection (Roslyn/Mono) is performed by specifying the module name.

Module scanning takes up the [lions share](https://www.dropbox.com/s/3y85wtznzio5sox/Screenshot%202015-03-07%2010.58.01.png?dl=0) of start up time.

If we special case the Roslyn and Mono modules for the automatic engine selection, we only have to scan for modules when someone specifies -m. This should have a dramatic speed increase when running scripts with specifying modules.

filipw:
Fixes #964 

Great perf benefit - improves startup from ~1.5s to ~0.5s.

Tested on Win and OS X

### DIFF
--- a/src/ScriptCs.Hosting/ModuleLoader.cs
+++ b/src/ScriptCs.Hosting/ModuleLoader.cs
@@ -12,7 +12,7 @@ namespace ScriptCs.Hosting
 {
     public class ModuleLoader : IModuleLoader
     {
-        private static readonly Dictionary<string, string> DefaultCSharpModules = new Dictionary<string, string>
+        internal static readonly Dictionary<string, string> DefaultCSharpModules = new Dictionary<string, string>
         {
             {"roslyn", "ScriptCs.Engine.Roslyn.dll"},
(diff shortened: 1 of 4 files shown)


adamralph:
Awesome @filipw!

Although the docker based Travis CI config is a vast improvement, it still fails with the intermittent xunit output path issue :unamused:. Re-running...

adamralph:
Added https://github.com/scriptcs/scriptcs/issues/964 to the release notes

adamralph:
I've tested locally and my total time to start REPL in an empty folder has gone from ~2 seconds to ~1 second :fistbump:

filipw:
I wonder why is that slow for you though, you have seen the profiler results (against debug build and started from VS) and it's ~500ms. And that's in a VM!

filipw:
:shipit:

PR #969, opened 2015-03-15
reverted AssemblyReferences changes since 0.13.x

### LINKED ISSUE #953: Duplicate references cause script execution to fail

adamralph:
reverts changes made in 61a8475da245653f7cd59287c5127afc15017bc3 and be4783577bd83dda8fb168a15717a3aba14c39df

closes #953

### DIFF
--- a/build/ScriptCs.Common.props
+++ b/build/ScriptCs.Common.props
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <OutputPath Condition=" '$(OutputPath)' == '' ">bin\$(Configuration)\</OutputPath>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <WarningsNotAsErrors>612,618</WarningsNotAsErrors>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
     <DebugSymbols>true</DebugSymbols>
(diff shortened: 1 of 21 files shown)


adamralph:
@filipw do you want this? It reverts the breaking changes to contracts which caused you problems with the new engine work.

adamralph:
Discussed with @filipw and we decided to take the hit of the breaking change in 0.14 since it's low impact and does tidy up things a lot.

I'll leave the branch dangling for a while just in case there's a change of heart so we can reopen.

Issue #956, opened 2015-03-05
Ignore shebang on *nix platforms


vtsingaras:
Make scriptcs don't treat shebang line as preprocessor directive.

mbaker-dev:
I'll take it!

Note: This is pretty similar to #675 (My work in progress on that issue already included this, so this may just be able to be closed as a duplicate?).

adamralph:
Awesome @mbaker-dev!

I think they are two separate features. One is to supply a bin stub. The other is to allow scripts to contain a shebang?

Issue #716, opened 2014-05-27
Homebrew Package


BrendanThompson:
Hey guys,

Would be super swell if we could get a hombrew brew for scriptcs :)

spboyer:
+1

filipw:
https://gist.github.com/filipw/6cbc6394360142a1976e

spboyer:
@filipw :+1:

c0ntinuum:
Hey Guys,

Just as an FYI here's one I built (based on a Gist by [akunzai](https://github.com/akunzai)):

https://github.com/c0ntinuum/homebrew-scriptcs

I'm happy to maintain this for the scriptcs project or you could just take it and make it part of a scriptcs or an associated repo.

Thanks for scriptcs ;)

filipw:
oh great, can we get it in here https://github.com/Homebrew/homebrew/tree/master/Library/Formula?

c0ntinuum:
@filipw I'll contact the Homebrew guys and see if we can get it in there ;)

adamralph:
Note that since the merge of #970 we have a binstub in the artifacts.

c0ntinuum:
@filipw scriptcs easily passes all criteria for inclusion in the primary Homebrew Formulae repository except the last one:

> has a stable, tagged version (i.e. not just a GitHub repository with no versions). See Interesting-Taps-&-Branches for where pre-release and head-only versions belong.

https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Formula-Cookbook.md#basic-instructions

> Formulae in the core repository must have a stable version tagged by the upstream project. Tarballs are preferred to git checkouts, and tarballs should include the version in the filename whenever possible.

> Software that does not provide a stable, tagged version, or had guidance to always install the most recent version, should be put in Homebrew/homebrew-head-only or homebrew/devel-only.

https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Acceptable-Formulae.md#stable-versions

Please let me know how you'd like me to proceed? I'm happy to raise an issue against Homebrew and see if they will take PR for the inclusion of the scriptcs.rb but it may well be a bit too early for that right now.

adamralph:
@c0ntinuum we do tag all our releases in GitHub, e.g. https://github.com/scriptcs/scriptcs/tree/v0.13.3

adamralph:
[...]

PR #970, opened 2015-03-16
add binstub and allow shebang

### LINKED ISSUE #675: Include binstub for *nix in build artifacts
via @paulcbetts https://gist.github.com/paulcbetts/e8514a03c6c01d1677cb

```bash
#!/usr/bin/env mono
## Make sure to chmod +x this file!

DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"

mono "$DIR/scriptcs.exe" $@
```

adamralph:
fixes #675 and #956

Raised on behalf of @mbaker-dev, even if just for discussion/review.


scriptcs
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+## Make sure to chmod +x this file!
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+mono "$DIR/scriptcs.exe" $@
```
adamralph: I think this should probably live under `/src/ScriptCs` with `Copy to Output Directory` set to `Copy Always`, so it just gets emitted into `bin/Release` along with everything else.

adamralph:
I think this looks great! Perhaps we should just tidy up the location of the binstub and then I'd say it's good to merge!

mbaker-dev:
I've moved the binstub to src/ScriptCs and included it in the csproj with CopyToOutputDirectory=Always. It's including the file in the build output now, as well as the artifacts area on Windows and *nix; however, it's not being included in the NuGet package. If it's included in the NuGet package, I think it should be specifically excluded from the Chocolatey install?

adamralph:
@mbaker-dev thanks again. Just having the binstub in the artifacts is a step forward. At the moment we don't provide any packages for Linux or Mac, but I've made a note on #716 and #972 that we now have the binstub in the artifacts.

adamralph:
I've added #675 and #956 to the v0.14 release notes.

BrendanThompson:
This is getting exciting!

PR #980, opened 2015-03-21
Only use fast path when extension is CSX or none

### LINKED ISSUE #964: Only scan for modules when -m is specified

filipw:
Fixes #964 and unblocks 0.14 release

The fast path is only invoked if there is:

 - no explicit module provided AND
 - file extension is "csx" OR no extension (REPL mode)

### DIFF
--- a/src/ScriptCs.Hosting/ModuleLoader.cs
+++ b/src/ScriptCs.Hosting/ModuleLoader.cs
@@ -72,7 +72,7 @@ namespace ScriptCs.Hosting
             if (modulePackagesPaths == null) return;
 
             // only CSharp module needed - use fast path
-            if (moduleNames.Length == 1 && DefaultCSharpModules.ContainsKey(moduleNames[0]) && (extension == DefaultCSharpExtension || string.IsNullOrWhiteSpace(extension))) 
+            if (moduleNames.Length == 1 && DefaultCSharpModules.ContainsKey(moduleNames[0]) && (string.IsNullOrWhiteSpace(extension) || extension.Equals(DefaultCSharpExtension, StringComparison.InvariantCultureIgnoreCase))) 
             {
                 _logger.Debug("Only CSharp module is needed - will skip module lookup");
                 var csharpModuleAssembly = DefaultCSharpModules[moduleNames[0]];
--- a/src/ScriptCs.Hosting/ModuleLoader.cs
+++ b/src/ScriptCs.Hosting/ModuleLoader.cs
@@ -18,6 +18,8 @@ namespace ScriptCs.Hosting
             {"mono", "ScriptCs.Engine.Mono.dll"}
         };
 
+        internal static readonly string DefaultCSharpExtension = ".csx";
+
         private readonly IAssemblyResolver _resolver;
         private readonly ILog _logger;
         private readonly Action<Assembly, AggregateCatalog> _addToCatalog;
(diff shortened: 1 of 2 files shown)


src/ScriptCs.Hosting/ModuleLoader.cs
```diff
@@ -69,7 +71,8 @@ public ModuleLoader(IAssemblyResolver resolver, ILog logger, Action<Assembly, Ag
         {
             if (modulePackagesPaths == null) return;
 
-            if (moduleNames.Length == 1 && DefaultCSharpModules.ContainsKey(moduleNames[0])) // only CSharp module needed
+            // only CSharp module needed - use fast path
+            if (moduleNames.Length == 1 && DefaultCSharpModules.ContainsKey(moduleNames[0]) && (extension == DefaultCSharpExtension || string.IsNullOrWhiteSpace(extension)))
```
adamralph: Should we make the comparison with `".csx"` case insensitive?

filipw: good point

adamralph:
Whilst we're here, should we also instantiate `DefaultCSharpModules` as case insenstive? I.e.

```C#
internal static readonly Dictionary<string, string> DefaultCSharpModules =
    new Dictionary<string, string>(StringComparer.CurrentCultureIgnoreCase)
    {
        {"roslyn", "ScriptCs.Engine.Roslyn.dll"},
        {"mono", "ScriptCs.Engine.Mono.dll"}
    };
```

src/ScriptCs.Hosting/ModuleLoader.cs
```diff
@@ -69,7 +71,8 @@ public ModuleLoader(IAssemblyResolver resolver, ILog logger, Action<Assembly, Ag
         {
             if (modulePackagesPaths == null) return;
 
-            if (moduleNames.Length == 1 && DefaultCSharpModules.ContainsKey(moduleNames[0])) // only CSharp module needed
+            // only CSharp module needed - use fast path
+            if (moduleNames.Length == 1 && DefaultCSharpModules.ContainsKey(moduleNames[0]) && (extension.Equals(DefaultCSharpExtension, StringComparison.InvariantCultureIgnoreCase) || string.IsNullOrWhiteSpace(extension)))
```
adamralph: This blows up if the extension is null (REPL). You'll have to use `string.Equals(extension, DefaultCSharpExtension, StringComparison.OrdinalIgnoreCase)`

adamralph: Or you could swap the order of the conditions around the `||`, but may as well just call the static on `string`.

adamralph:
The Travis build is still so flaky... IIRC I've seen the problem before. It's a race condition in xunit 1.0. I don't know why it only manifests in Linux though.

adamralph:
LGTM. The module name casing can be left for now, that's a wider problem which isn't catered for in `ScriptServicesBuilder` either (and is unlikely to be affecting anyone anyway).

Issue #947, opened 2015-03-04
0.14 release


adamralph:
~~- [ ] revert breaking changes to AssemblyReferences in #953 (PR raised https://github.com/scriptcs/scriptcs/pull/969) (@adamralph)~~
- [x] fix regression in #964 (@filipw + @adamralph)
- [x] revert breaking ScriptExecutor change in #909 (PR https://github.com/scriptcs/scriptcs/pull/981) - [x] ensure all 0.14 issues are attached to the milestone and labelled appropriately (@adamralph)
(@adamralph + @glennblock)
- [x] bump version number (PR https://github.com/scriptcs/scriptcs/pull/95) (@adamralph + @khellang)
- [x] draft GitHub release *with tag on* ***master*** (@adamralph)
- [x] smoke tests
  - [x] script packs (@adamralph)
    - [x] Bau
    - [x] Nancy
  - [x] modules
    - [x] gist
    - [x] CSharp (@filipw)
    - [x] VisualBasic (@filipw)
  - [x] hosts (@adamralph)
    - [x] ConfigR
- [x] ensure PR for blog post regarding #909 is ready to be merged (PR https://github.com/scriptcs/scriptcs.github.com/pull/2) (@paulbouwer + @glennblock)
- [x] raise and merge PR from dev to master
- [x] confirm packages have been uploaded to Chocolatey
- [x] publish GitHub release
- [x] merge blog post PR
- [x] announce on official Twitter account
- [x] raise PR for version bump

glennblock:
Marked as hold based on breaking change related to module loading. @filipw will fix

filipw:
#968

glennblock:
Also added module loading tests as a prereq for the release

filipw:
:shipit:

filipw:
both of the new engine modules work with this codebase, if that counts for a smoke test

adamralph:
Ah, indeed it does!

PR #984, opened 2015-03-23
change version from 0.14 to 0.15

adamralph:
re #947

We need to start doing this immediately after each release, otherwise it's impossible for a consuming project which is currently using our 0.14 NuGet packages to update to our nightly NuGet packages.


glennblock:
:+1:

PR #986, opened 2015-03-23
bye bye Common.Logging, hello LibLog...

### LINKED ISSUE #847: Deprecate Common.Logging usage in favour of our own logging API

adamralph:
fixes #847


glennblock:
Can't wait for this one to pass!

adamralph:
Argghhh LibLog induces a mono compiler warning - and it's a source package.

Guess I'll just alter it. We'll just have to be careful when we upgrade to a newer version.

adamralph:
I think the mono compiler is telling big lies...

adamralph:
Hopefully that should do it. I raised https://github.com/damianh/LibLog/issues/46 but I doubt there's anything Damian can do about it.

adamralph:
OK, that's fixed the compiler warning. Now Travis is simply acting the goat as usual...

adamralph:
Hooray! Green at last!

I only had to hammer the rebuild button in Travis about 7 or 8 times...

glennblock:
OMG 80 files. Was this manual or search and replace?

Settings.StyleCop
```diff
@@ -6,6 +6,15 @@
       <Value>nullable</Value>
     </CollectionProperty>
   </GlobalSettings>
+  <Parsers>
+    <Parser ParserId="StyleCop.CSharp.CsParser">
```
glennblock: Why is this here?

src/ScriptCs.Core/App_Packages/LibLog.3.1/LibLog.cs
```diff
@@ -0,0 +1,1941 @@
+//===============================================================================
```
glennblock: 1900 lines of code in one file really? Is there no refactored version? We can have a logging folder and put individual files there.

adamralph:
It was manual

Settings.StyleCop
adamralph: This stops StyleCop from analysing the LibLog file.

glennblock:
I bit my tongue and lip and pressed merge.

jrusbatch:
:clap:

glennblock:
Thanks for putting in what must have been an incredibly boring, but useful effort :+1:

adamralph:
I quite enjoyed it :smile:. I'm a bit special like that

ztone:
Nice work @adamralph :tada:

Settings.StyleCop
damianh: In ur codez, pollutin' ur stylez

src/ScriptCs.Core/App_Packages/LibLog.3.1/LibLog.cs
damianh: I made a conscious decision to ship as a single file. Don't want to crap stuff all over peoples' csprojs. Also helps reduce chances of accidental editing / refactoring. If you _do_ edit it, an update to a newer version won't work (smoothly). In my experience of using source code packages, I've found life is easier this way.

While it's also developed as a single file but even if I split that out I'd still ship as a single file.

damianh: Oh and TinyIoC does the same thing for similar reasons.

glennblock: @damianh I was initially against the idea, but as I reflected on it more, and the benefits like you cited I warmed up to it. Common.Logging needed to go, so this is a great replacement.

Thanks!

filipw:
https://github.com/damianh/LibLog/releases/tag/v4.0.0

damianh:
I've delisted 4.0.0 pending this https://github.com/damianh/LibLog/issues/48 Would appreciate your input on this. Would like to re-publish 4 as quickly as possible.

Issue #998, opened 2015-03-26
Add LoadScriptPacks method to ScriptServicesBuilder


glennblock:
When hosting you have to currently set ScriptName to empty string or set Repl to true in order to get script packs to load.

This is non-intuitive and has thrown off hosters like @laedit.

filipw:
tbh, it has thrown off hosters like @filipw too

glennblock:
Uh yeah and it threw me off a few times and I wrote the builder #facepalm

glennblock:
Fortunately this is a REALLY simple fix

glennblock:
Working on this now. @laedit stay tuned.

laedit:
Excellent, thanks! :)

glennblock:
ping @laedit

filipw:
this is not breaking api though

adamralph:
`IScriptServicesBuilder` has had a member added to it, so anyone deriving from it will break.

filipw:
my point was this is really private scriptcs API since there is no way to plug it yourself or really do anything with it. in fact the interface is probably not needed at all

adamralph:
Fair enough - we should remove it then. But that is also, strictly speaking, a breaking change :wink:

filipw:
yes you are right

glennblock:
The interface is for unit testing.

glennblock:
I think the impact here is low. I could use an extension method and have it
do some casting magic under the hood, bur I don't think that is necessary.

This should not break hosters implementations. What could break is their
unit tests if they have handrolled mocks.

So technically a breaking change but practically not.

adamralph:
Agreed. Extremely low impact. We can just :shipit:

filipw:
imho we should change interface it to just the builder class with virtuals (as it's just for testing) - one less code file to maintain

glennblock:
That is true. This rarely changes, but yes a consideration. Is

PR #999, opened 2015-03-27
Adding LoadScriptPacks to builder and tests for builder script pack load...

glennblock:
...ing

Fix for #998.

@laedit please review this.

### DIFF
--- a/src/ScriptCs.Hosting/ScriptServicesBuilder.cs
+++ b/src/ScriptCs.Hosting/ScriptServicesBuilder.cs
@@ -20,7 +20,7 @@ namespace ScriptCs.Hosting
         private Type _scriptExecutorType;
         private Type _replType;
         private Type _scriptEngineType;
-        private bool _loadScriptPacks;
+        private Nullable<bool> _loadScriptPacks;
 
         public ScriptServicesBuilder(
             IConsole console,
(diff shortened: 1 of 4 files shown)


laedit:
Seems perfect! :smiley:

src/ScriptCs.Hosting/ScriptServicesBuilder.cs
```diff
@@ -48,7 +49,7 @@ public ScriptServices Build()
 
             _scriptEngineType = (Type)Overrides[typeof(IScriptEngine)];
 
-            var initDirectoryCatalog = _scriptName != null || _repl;
+            var initDirectoryCatalog = _scriptName != null || _repl || _loadScriptPacks;
```
filipw: this is still kinda confusing since I can say `LoadScriptPacks(false)` and yet it will override me because I have set a filename. It would be better to just check `_loadScriptPacks` here - this way I can have script execution or REPL without initializing script packs if I don't care about them in my host

glennblock: Good point

glennblock: I am not sure I buy this yet. If I do what you are saying it will break all hosters today that use script packs.

glennblock: I have an idea using Nullable<bool>, this will allow me to know whether Load was set explicitly or not.

glennblock:
OK @filipw take a look. Using a `Nullable<bool>` I was able to preserve the former behavior, but now allow you to explicitly disable / enable loading.

filipw:
yeah - that's good, thanks

glennblock:
@filpw thanks for catching this, I agree they should not be coupled.

glennblock:
@filipw can you press the merge button? :-)

filipw:
:thumbsup:

glennblock:
Thanks man! It's great the way we help each other :-)

PR #1000, opened 2015-03-27
Support for namespace aliasing

### LINKED ISSUE #826: Namespace aliasing causes an exception to be thrown
Hi there,

By default, System.IO namespace is added to every script.

Problem is I have an assembly that defines Fluent.IO.Path that conflicts with System.IO.Path.

So, I'd need to write something like this:

using Path=Fluent.IO.Path;

However, when I try to run my script, I get the following error:
"error CS0246: The type or namespace name 'Path' could not be found (are you missing a using directive or an assembly reference?)"

Any help would be great.

Running scriptcs in debug mode, I can confirm the assembly is being found and loaded.

Thank you,
Brian

https://groups.google.com/forum/m/?fromgroups#!topic/scriptcs/7YiOeeGgaZw

filipw:
Currently we don't allow namespace aliasing - this PR fixes it.

Fixes #826

### DIFF
--- a/src/ScriptCs.Core/UsingLineProcessor.cs
+++ b/src/ScriptCs.Core/UsingLineProcessor.cs
@@ -30,7 +30,7 @@ namespace ScriptCs
 
         private static bool IsUsingLine(string line)
         {
-            return line.Trim(' ').StartsWith(UsingString) && !line.Contains("{") && line.Contains(";");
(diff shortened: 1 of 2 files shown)


filipw:
aliases are forced to be treated as regular code here meaning they are executed together with script. Works for REPL too.

screenshot 2015-03-27 09 43 28

adamralph:
Issue/PR 1000! :tada:

adamralph:
OMG Travis is green. *Hurries to the merge button...*

glennblock:
:+1:

PR #1002, opened 2015-03-27
add issue stats badges to readme

adamralph:


filipw:
btw, why was this branch part of the upstream?

adamralph:
It's because I made the change using the GitHub UI. I don't think it has an
option to use your own fork for the branch.

khellang:
That's the same reason the version bump landed on an upstream branch. Didn't realize we were so careful about polluting upstream with temporary branches. IIRC, you can use the GH UI from the fork to do a PR. I'll remember to do that next time.

filipw:
I don't think it's that important to be honest but I think it's a nice practice since it looks much nicer to the outsiders.

Just compare Nancy (https://github.com/NancyFx/Nancy/branches/all) to MVC (https://github.com/aspnet/Mvc/branches/all) :smile:

adamralph:
:+1: In this case it's not so important, but generally I abhor upstream branches for this very reason.

Issue #1001, opened 2015-03-27
Upgrade to xunit 2 and xbehave 2


adamralph:
Because xunit 1 is just so 2013 :stuck_out_tongue_winking_eye: 

Benefits:
- xunit 2.0
  - This will _probably_ get rid the intermittent (9/10) build failures which we see in Travis CI, which I believe is a problem which only manifests in xunit 1.x, on low spec VMs (some kind race condition)
  - parallel test exec
  - self contained VS runner support
  - test collections
  - nicer console output
  - other new xunit 2.0 features - http://xunit.github.io/release-notes/2015-03-16.html
- xbehave 2.0
  - support for all new xunit 2.0 features
  - all steps following a failed step in a scenario are reported as skipped instead of failed, drastically reducing noise in test output
  - other minor changes - https://github.com/xbehave/xbehave.net/wiki/Changes-in-version-2.0
- moving off obsolete platforms which are purely in maintenance mode

Blockers:
- xunit 2.0 currently does not work in Mono, due to a bug in Mono, but this has been fixed and is awaiting release. See https://github.com/xunit/xunit/issues/158
- AutoFixture.Xunit is still tied to xunit 1.x - see https://github.com/AutoFixture/AutoFixture/pull/294
- xbehave 2.0 is still in RC, but RTM should be in the next week or two - see https://github.com/xbehave/xbehave.net/issues/144

adamralph:
Note that we can probably spin off an issue to upgrade the acceptance tests first, since they are not blocked by AutoFixture,Xunit, and that in itself should fix Travis since, at the moment, only the acceptance tests are run there.

adamralph:
But of course, an update to the acceptance tests is still dependent on the Mono fix.

filipw:
Fixed by #1194

PR #1005, opened 2015-03-28
Added :scriptpacks REPL command

filipw:
Allows to easily inspect the script packs available in the REPL context.

 - prints method signatures (including extension methods)
 - prints property definitions (including correct getter/setter)
 - uses fully qualified type names if the given namespace is not imported into REPL context, if it is imported uses simple type name

Example:
screenshot 2015-03-28 23 31 51

### DIFF
--- a/src/ScriptCs.Core/ReplCommands/ScriptPacksCommand.cs
+++ b/src/ScriptCs.Core/ReplCommands/ScriptPacksCommand.cs
@@ -44,8 +44,7 @@ namespace ScriptCs.ReplCommands
                 var contextType = packContext.GetType();
                 _console.WriteLine(contextType.ToString());
 
-                var methods = contextType.GetMethods(BindingFlags.Instance | BindingFlags.Public | BindingFlags.DeclaredOnly).
-                    Where(m => !m.IsSpecialName).Union(contextType.GetExtensionMethods()).OrderBy(x => x.Name);
+                var methods = contextType.GetAllMethods();
                 var properties = contextType.GetProperties(BindingFlags.Instance | BindingFlags.Public | BindingFlags.DeclaredOnly);
 
                 PrintMethods(methods, importedNamespaces);
@@ -59,7 +58,7 @@ namespace ScriptCs.ReplCommands
 
         private void PrintMethods(IEnumerable<MethodInfo> methods, string[] importedNamespaces)
         {
-            if (methods.Any())
+            if (!methods.IsNullOrEmpty())
             {
                 _console.WriteLine("** Methods **");
                 foreach (var method in methods)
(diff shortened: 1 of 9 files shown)


filipw:
why do we even have travis if it fails all the time ;)

adamralph:
It passes in about 1 run of out 10 :wink:

Hopefully this will get rid of the problem https://github.com/scriptcs/scriptcs/issues/1001

glennblock:
@filipw I share your pain. We need to do something and hopefully #1001
fixes. If not, we need to make other plans.

filipw:
anyone?

adamralph:
My last few days have been chaotic, but I can look at the weekend unless someone beats me to it.

src/ScriptCs.Core/ReplCommands/ScriptPacksCommand.cs
```diff
@@ -0,0 +1,185 @@
...
+        public string CommandName
+        {
+            get { return "scriptpacks"; }
+        }
+
+        public object Execute(IRepl repl, object[] args)
+        {
+            var packContexts = repl.ScriptPackSession.Contexts;
+            var originalColor = _console.ForegroundColor;
+            _console.ForegroundColor = ConsoleColor.Yellow;
```
adamralph: This line should be moved down to inside the conditional, just before writing out the message to say there are no script packs. (See further comment below).

src/ScriptCs.Core/ReplCommands/ScriptPacksCommand.cs
```diff
@@ -0,0 +1,185 @@
...
+            }
+
+            var importedNamespaces = repl.ScriptPackSession.Namespaces.IsNullOrEmpty() ? repl.Namespaces.ToArray() : repl.Namespaces.Union(repl.ScriptPackSession.Namespaces).ToArray();
+
+            foreach (var packContext in packContexts)
+            {
+                var contextType = packContext.GetType();
+                
+                _console.WriteLine(contextType.ToString());
+                _console.ForegroundColor = originalColor;
```
adamralph: Because the yellow colour is set outside this loop, the first script pack name will be printed in yellow, but every subsequent one will be printed in green. The yellow should be set just before writing out the name of the script pack.

Also, since this a static state mutation, after setting the yellow, the `WriteLine` should really be a in a `try{}` with the resetting of the colour in a finally.

(Once I pull my finger out and produce a source package for https://github.com/colored-console/colored-console, we can use it to wrap up colour management for us.)

filipw: you are right, i think i messed it up at some point (look at the screenshot, it was ok)

adamralph:
nice @filipw, I love this new command

adamralph:
I've given it pride of place in the 0.15 release notes :smile:.

Issue #996, opened 2015-03-25
0.15 release


adamralph:
- [x] decide how to proceed wrt Common.Logging
- [x] close ~~#847~~
  - [x] merge ~~#1009~~
  - [x] re-introduce Common.Logging overloads, marked as obsolete - ~~https://github.com/scriptcs/scriptcs/pull/1068~~ ~~https://github.com/scriptcs/scriptcs/pull/1071~~
- [x] ensure any patches have been merged from master to dev
- [x] ensure all issues are attached to the milestone and labelled appropriately
- [x] draft GitHub release *with tag on* ***master***, including details of breaking changes and impact
- [x] smoke tests
  - [x] script packs
    - [x] Working samples
      - [x] Bau
      - [x] Nancy
    - [x] [Script Pack Smoke Tester](https://github.com/adamralph/scriptcs-testing-script-packs)
      - [x] WebApi2 ***FAILS!*** insists on Common.Logging.ILog being injected. Will have to be fixed and released after the 0.15 release - raised https://github.com/scriptcs-contrib/scriptcs-webapi/issues/20
      - [x] the rest
  - [x] modules
    - [x] gist
    - [x] CSharp
    - [x] VisualBasic
  - [x] hosts
    - [x] ConfigR
- [x] raise and merge PR from dev to master - ~~#1074~~
- [x] confirm packages have been uploaded to Chocolatey
- [x] publish blog post - ~~https://github.com/scriptcs/scriptcs.github.com/pull/5~~
- [x] publish GitHub release (after updating blog post URL)
- [x] raise SVM PR - https://github.com/scriptcs-contrib/svm/pull/75
- [x] get sha256 of tar.gz file from release and raise homebrew PR - https://github.com/Homebrew/homebrew/pull/42417
- [x] announce on official Twitter account
- [x] raise PR for version bump

adamralph:
Currently we have one major API feature ready to go (removal of Common.Logging) plus two minor exe features and two minor bugs - https://github.com/scriptcs/scriptcs/issues?q=milestone%3Av0.15+is%3Aclosed

I'd also like to see https://github.com/scriptcs/scriptcs/pull/944 make it into this release.

adamralph:
So, with the aid of https://github.com/adamralph/scriptcs-testing-script-packs I've ascertained that only one script pack is broken by 0.15 - ScriptCs.WebApi2.

[...]

PR #1009, opened 2015-04-01
close 847, solve 1008

### LINKED ISSUE #847: Deprecate Common.Logging usage in favour of our own logging API

### LINKED ISSUE #1008: Show logger name at debug and trace level
e.g.

`DEBUG: [ScriptCs.SomeNamespace.SomeType] Some message`

### LINKED ISSUE #1010: Better console output when script not found
```
C:\code>scriptcs awol.csx
ERROR: Could not find file 'C:\code\awol.csx'. - 'C:\code\awol.csx'.

C:\code>
```

Silly.

adamralph:
Re #847 - move the logging interfaces from Core to Contracts, so that anyone referencing Contracts can continue to get one injected, as the WebApi2 script pack does

fixes #1008 by switching from passing around `ILog` to passing around an `ILogProvider`. This also allows anyone who gets an logging interface injected to create a logger specific to them. E.g. the WebApi2 script pack can create a "WebApiPack" logger or even one for each type. If we expose `ILogProvider` to scripts, they can create a logger with the script name, etc.

This is the console output at debug level after this change:

image

fixes #1010


adamralph:
Since we are breaking the API with respect to logging in 0.15, I figured we may as well take advantage of it and switch from `ILog` to `ILogProvider at the same time, so we can do things like #1008 and also allow scripts, script packs, modules etc. create their own named loggers.

Note that this does not detract in anyway the options listed at https://github.com/scriptcs/scriptcs/issues/996#issuecomment-87377620. If we decide to go with option 3, i.e. a deprecation of Common.Logging by maintaining the Common.Logging overloads, we can still do that, by adding them back in, which we leave us with both `Common.Logging.Ilog` and `ScriptCs.Contracts.ILogProvider` overloads.

adamralph:
Trying to figure out what happened with the CI build.

Works On My Machine™

adamralph:
That turned out to be rather subtle. The tests were all passing in VS with whatever test runner I tried, but were failing on the command line. I then switched VS to release config and it also failed. I tracked it down to this https://github.com/adamralph/scriptcs/blob/871c88bee11ec0130e9383eee52f26a2a4552960/src/ScriptCs.Contracts/Logging/LogProviderExtensions.cs#L14

The JIT was inlining the method in the release build which fatally altered the stack, causing this particular piece of magic to fail. It's the same method as used in LibLog so it's proven to work (https://github.com/damianh/LibLog/blob/69873d63d382073ac16266a071654fb01fdd4b46/src/LibLog/LibLog.cs#L438-L439) but over there it's not an extension method so it doesn't get inlined. I tried splitting it into two lines to match the LibLog impl. exactly but it made no difference.

I guess LibLog may also be liable to this problem, but it's been getting lucky so far just because of the way that the JIT heuristics work out. I'll probably raise an issue over there for consideration.

khellang:
You could try adding a `MethodImpl` attribute with `MethodImplOptions.NoInlining`. See https://msdn.microsoft.com/en-us/library/system.runtime.compilerservices.methodimploptions(v=vs.110).aspx

khellang:
Uh. Looks like that's already done. Damn mobile views...

adamralph:
I added a commit to fix https://github.com/scriptcs/scriptcs/issues/1010

Before:

```
C:\code>scriptcs awol.csx
ERROR: Could not find file 'C:\code\awol.csx'. - 'C:\code\awol.csx'.

C:\code>
```

After:
```
C:\code>scriptcs awol.csx
ERROR: Error executing script 'awol.csx' [FileNotFoundException] Could not find file 'C:\code\awol.csx'.

C:\code>
```

adamralph:
As far as I can see, this PR now completes all the logging changes envisioned so far.

It is using the latest version of LibLog (4.1.1) which contains a new feature which I contributed (https://github.com/damianh/LibLog/issues/56) allowing LibLog to function purely as a provider of out of the box support for common logging frameworks and sitting, internally, underneath the public logging API of the consuming library.

To summarise:

* ScriptCs.Contracts defines it's own public logging API, decoupled from LibLog, with the initial definition copied from LibLog (because it's very good).
* All types in other projects now accept `ILogProvider` parameters instead of `ILog` to allow specific loggers to be created to support features such as https://github.com/scriptcs/scriptcs/issues/1008
* ScriptCs.Contracts provides a DefaultLogProvider which uses LibLog providers internally to provide out of the box support for common logging frameworks. This is not used by scriptcs.exe. It is purely for hosts.
* ScriptCs.Contracts provides a NullLogProvider as a cheap way for hosts to disable logging without having to install a logging framework and provide a null config (LibLog logs to the console by default)

To clarify, these changes are independent of the decision we take regarding breaking changes for 0.15 (see https://github.com/scriptcs/scriptcs/issues/996#issuecomment-87377620). If we do decide to reinstate the Common.Logging overloads for a transition period then this can be done later, independently of these changes.

adamralph:
rebased

adamralph:
Added one more commit: LibLog 4.2 no longer logs to the console by default which removes the need to publicly expose `NullLogProvider` for hosts to switch this off. They can just use `DefaultLogProvider` in all cases.

adamralph:
Rebased and upgraded to LibLog 4.2.2

adamralph:
rebased

filipw:
LGTM

PR #1014, opened 2015-04-06
Add missing null checks found by static analysis

scottcarr:
Hi, these are some more missing null checks my static analyzer found.

Note: the analyzer didn't suggest a null check for lineProcessors in FilePreProcessor.  I'm not sure why.  I can add it if you want.

### DIFF
--- a/src/ScriptCs.Core/FilePreProcessor.cs
+++ b/src/ScriptCs.Core/FilePreProcessor.cs
@@ -17,6 +17,9 @@ namespace ScriptCs
 
         public FilePreProcessor(IFileSystem fileSystem, ILog logger, IEnumerable<ILineProcessor> lineProcessors)
         {
+            Guard.AgainstNullArgument("fileSystem", fileSystem);
+            Guard.AgainstNullArgument("logger", logger);
+
             _fileSystem = fileSystem;
             _logger = logger;
             _lineProcessors = lineProcessors;
--- a/src/ScriptCs.Core/LoadLineProcessor.cs
+++ b/src/ScriptCs.Core/LoadLineProcessor.cs
@@ -14,6 +14,8 @@ namespace ScriptCs
 
         public LoadLineProcessor(IFileSystem fileSystem)
         {
+            Guard.AgainstNullArgument("fileSystem", fileSystem);
+
             _fileSystem = fileSystem;
         }
 
--- a/src/ScriptCs.Core/ReferenceLineProcessor.cs
+++ b/src/ScriptCs.Core/ReferenceLineProcessor.cs
@@ -14,6 +14,8 @@ namespace ScriptCs
 
         public ReferenceLineProcessor(IFileSystem fileSystem)
         {
+            Guard.AgainstNullArgument("fileSystem", fileSystem);
+
             _fileSystem = fileSystem;
         }
 
--- a/src/ScriptCs.Core/ReplCommands/AliasCommand.cs
+++ b/src/ScriptCs.Core/ReplCommands/AliasCommand.cs
@@ -12,6 +12,8 @@ namespace ScriptCs.ReplCommands
 
         public AliasCommand(IConsole console)
         {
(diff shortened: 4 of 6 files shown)


adamralph:
@scottcarr thanks!

This is definitely one area that needed addressing. We're fairly well covered in methods that dereference their arguments but the case of 'assign field from argument in constructor and dereference later' we definitely had gaps and I'm sure that more remain. Unless you think you've caught them all now?

scottcarr:
There are more.  The tool is conservative -- meaning it won't necessarily suggest every single contract, but the contracts it does suggest should never break your code.

Also, I've been hand curating the suggested contracts.  For example: [this line](https://github.com/scriptcs/scriptcs/blob/dev/src/ScriptCs.Core/AppDomainAssemblyResolver.cs#L94) confused the static analyzer.  It thinks foundInfo.Assembly might be a null dereference, but it's not -- because if TryGetValue returns false (the case where foundInfo will be null) the if clause short circuits (I think).  It took me several minutes of thinking to realize this.

The other reason I've been doing manual filtering is the tool doesn't recognize your existing contracts (because the tool looks for, and suggests, [CodeContracts](http://research.microsoft.com/en-us/projects/contracts/) contracts).  So if the code has a Guard.AgainstNullArgument(foo) the tool doesn't realize you've already got a null check and will suggest adding a duplicate.  It is smart enough to realize what to do if you explicitly handle the null case though.  For example it (correctly) won't suggest a contract for something like this:

```
class Foo 
{
    Bar _bar;
    ...
    void foo(Bar bar) 
    {
       if (bar == null) 
      { 
        bar = new Bar();
      } 
      _bar = bar;
    }
}
```  

In summary: there are more missing checks. I just need to comb through the suggestions.  There is also the possibility that the tool could discover more missing contracts if all your existing contracts were converted to CodeContracts contracts (or somehow  your contracts were annotated to make the tool aware of them) but that would be a bit more effort!  I will investigate this after I look through the suggestions one more time.

glennblock:
Thanks for the effort! I would say using Code Contracts is definitely not a
path we want to go down.

scottcarr:
@glennblock Yeah, I totally understand.  The good news is for some contracts like (null checks) there's an obvious and easy conversion from Code Contracts to your contracts.

adamralph:
Yeah, Code Contracts as in, [Code Contracts](https://msdn.microsoft.com/en-us/library/dd264808%28v=vs.110%29.aspx?f=255&MSPPError=-2147217396) is, IMHO, a half baked solution and I'm not a fan of it.

Static analysis can only ever prove so much but it certainly helps. @scottcarr is this something you wrote yourself? If you want to make it more intelligent, you could check usages of [`ValidatedNotNullAttribute`](https://github.com/scriptcs/scriptcs/blob/9a3b4b6f1d83086a5558b874d6b265cf2a3ffa6f/src/ScriptCs.Core/Guard.cs#L109), e.g. [here](https://github.com/scriptcs/scriptcs/blob/9a3b4b6f1d83086a5558b874d6b265cf2a3ffa6f/src/ScriptCs.Core/Guard.cs#L23). This is what stops VS code analysis from showing false positives when `Guard` is used.

scottcarr:
@adamralph Yeah, I saw that attribute but wasn't sure what it did.  I didn't write the analyzer.  My tool takes the output of cccheck (a code contract analyzer) and automatically inserts the contracts using Roslyn.  Unfortunately it's all very closely coupled with [Code Contracts.](https://visualstudiogallery.msdn.microsoft.com/1ec7db13-3363-46c9-851f-1ce455f66970)  That's just an FYI, I'm not trying to sell you on it.

If you want to see every single contract the analysis suggests, this branch of my scriptcs fork has them all: https://github.com/scottcarr/scriptcs/commit/6732de6e7c7c7129a62bc771dbc81ffe12661563

adamralph:
Ah, I see so it's cccheck that is raising the false positives...

PR #1017, opened 2015-04-07
Add rest of the missing null checks (that I could find)

scottcarr:
This is rest I could find except one (potentially).

When I was looking [here](https://github.com/scriptcs/scriptcs/blob/dev/src/ScriptCs.Core/Repl.cs#29) I thought it was weird that the tool didn't suggest a null check for _scriptArgs.  I don't know why exactly, but I believe it has something to do with the parameter scriptEngine being an interface type.  scriptArgs only ever gets passed as the second argument to ScriptEngine.Execute.  Since IScriptEngine is a public interface I can't know all the possible implementations of it, right?  So I can't know if the implementer will just dereference the null parameter or check for it.

I looked through the code for the implementers of IScriptEngine and its only inherited by another interface (IReplEngine).  IReplEngine is only implemented by MonoScriptEngine and RoslynReplEngine.  They both pass the second argument of Execute to the ScriptPackManager constructor which does check for null!  That was a very long-winded way of saying there is no missing null check for the implementers in the existing code, but a different implementer might not ... and this is an example of why finding all the missing null checks is pretty complicated :)

### DIFF
--- a/src/ScriptCs.Core/Repl.cs
+++ b/src/ScriptCs.Core/Repl.cs
@@ -26,6 +26,9 @@ namespace ScriptCs
             IEnumerable<IReplCommand> replCommands)
             : base(fileSystem, filePreProcessor, scriptEngine, logger, composer)
         {
+            Guard.AgainstNullArgument("console", console);
+            Guard.AgainstNullArgument("serializer", serializer);
+            
             _scriptArgs = scriptArgs;
             _serializer = serializer;
             Console = console;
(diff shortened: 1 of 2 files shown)


adamralph:
Personally, I wouldn't add a null check for `scriptArgs` in the `Repl` constructor since the class doesn't dereference the assigned field. If an implementer method does, then it should be responsible for doing its own null check.

In a different case, if a type were encapsulating some concept and a field representing some aspect of that would make no sense with a null assignment then I would throw an exception before assigning the field, but I'm not sure we can really define such an invariant in this case.

adamralph:
Thanks again!

PR #1019, opened 2015-04-11
updated to Nuget.Core 2.8.5 and renamed "NuGet.Core" to "Nuget.Core"

filipw:
NuGet.Core was renamed to Nuget.Core recently.

Since we reference "NuGet.Core" it is impossible to install scriptcs packages using KPM because KPM is currently case sensitive.

@scriptcs/core this should be merged and shipped ASAP as it's preventing anyone to use scriptcs.hosting in ASP.NET 5


filipw:
Here's what happens when trying to install `ScriptCs.Hosting` into ASP.NET 5 right now

screenshot 2015-04-11 11 55 20

filipw:
green now after fixing paths

jrusbatch:
:+1:

Issue #1018, opened 2015-04-09
Use packageSourceCredentials when interacting with feeds


aarondandy:
While `scriptcs -I` does check other feeds for packages that are listed in my NuGet.Config it does not send any credentials in the requests that are found in the `packageSourceCredentials` element.

glennblock:
Good catch. With nuget supporting secure repos we should. Wanna take this?

aarondandy:
It looks like Chocolatey is "Doing It Right" so I may investigate how they do it first. I'll look into it a bit more before I commit to doing it as I really don't know enough about NuGet yet. This is important enough for me that I will probably end up learning more about NuGet though and obviously I have some feeds that require authentication that I can work with.

glennblock:
Fabulous! :-)

aarondandy:
I have a fork and am looking into this.

I was wrong about Chocolatey doing it right, those calls I was watching in Fiddler were directly from nuget.exe .

filipw:
This was fixed in 0.17.0

PR #1021, opened 2015-04-13
Set the global credential provider #1018

aarondandy:
Fix for #1018 that sets the global credential provider. This allows the NuGet HttpClient to use the credentials stored in NuGet.Config .

I tested this by editing my `%APPDATA%\NuGet\NuGet.Config` file to use `packageSourceCredentials` for our myget feeds, specifically our nuget mirror. This file has Username and Password credentials for some of the MyGet feeds and nuget.org is disabled using:

```xml
  <disabledPackageSources>
    <add key="https://www.nuget.org/api/v2/" value="true" />
  </disabledPackageSources>
```

I produced two builds using `build.cmd`, one from before the change and one from after and watched what they did using Fiddler. The build before the change would try the various feeds once and get a 401 response for each then just give up. The build after the change would retry with credentials following the 401 response getting 404s for most and a 200 for the package on the mirror feed. More importantly the build with the fix actually installed some packages too!

Regarding the approach, to my knowledge setting this global property is the only way to configure the NuGet HttpClient to use these credentials. It kind of makes me wonder if this should be fixed in NuGet.Core instead of here in script CS but I guess that is up to you all and those working on NuGet v2.

The location for this one line was chosen because of the proximity to the `PackageSourceProvider` instance. I would like to see this code in `Initialize` but I don't think setting it multiple times should hurt and I also don't see the `GetRepositorySources` method called anywhere except for `Initialize`.

Some things to note:

* No tests were created for this change. I really don't know a practical way to create a test for this change.
* The code that was edited does not appear to have any existing coverage (using NCrunch).
* I have no test failures before or after this change using `build.cmd`.

### DIFF
--- a/src/ScriptCs.Hosting/Package/NugetInstallationProvider.cs
+++ b/src/ScriptCs.Hosting/Package/NugetInstallationProvider.cs
@@ -55,6 +55,9 @@ namespace ScriptCs.Hosting.Package
             }
 
             var sourceProvider = new PackageSourceProvider(settings);
+
+            HttpClient.DefaultCredentialProvider = new SettingsCredentialProvider(NullCredentialProvider.Instance, sourceProvider);
+
             var sources = sourceProvider.LoadPackageSources().Where(i => i.IsEnabled == true);
 
             if (sources == null || !sources.Any())


aarondandy:
This may be an alternative way to solve this problem in the future if scriptcs can allow onfiguration for it: http://docs.nuget.org/Consume/Credential-Providers

glennblock:
@scriptcs/core thoughts?

glennblock:
This looks reasonable to me as long as it does not affect existing users that are not requiring credentials.

PR #1032, opened 2015-05-02
Format Help Command Output

BlackFrog1:
Word-wrap the output of each command description via the help command. The Help Command Test Unit was updated for the new layout.
 
Window Width @ 80 character - standard
Window Wisth @ 40 character - extreme example

### DIFF
--- a/src/ScriptCs.Core/App_Packages/LibLog.3.1/LibLog.cs
+++ b/src/ScriptCs.Core/App_Packages/LibLog.3.1/LibLog.cs
@@ -463,7 +463,7 @@ namespace ScriptCs.Logging
 #else
                 Console.WriteLine(
 #endif
-                    "Exception occurred resolving a log provider. Logging for this assembly {0} is disabled. {1}",
+                    "Exception occured resolving a log provider. Logging for this assembly {0} is disabled. {1}",
                     typeof(LogProvider).GetAssemblyPortable().FullName,
                     ex);
             }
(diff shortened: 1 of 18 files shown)


src/ScriptCs.Core/App_Packages/LibLog.3.1/LibLog.cs
```diff
@@ -1814,7 +1814,7 @@ internal static class LogMessageFormatter
         /// <summary>
         /// Some logging frameworks support structured logging, such as serilog. This will allow you to add names to structured data in a format string:
         /// For example: Log("Log message to {user}", user). This only works with serilog, but as the user of LibLog, you don't know if serilog is actually 
-        /// used. So, this class simulates that. it will replace any text in {curlybraces} with an index number. 
+        /// used. So, this class simulates that. it will replace any text in {curly braces} with an index number. 
         /// 
         /// "Log {message} to {user}" would turn into => "Log {0} to {1}". Then the format parameters are handled using regular .net string.Format.
         /// </summary>
```
adamralph: This file is imported by the LibLog package so we really shouldn't edit it. Any changes here should be submitted as a PR to https://github.com/damianh/LibLog

adamralph:
@BlackFrog1 this is a nice change, thanks!

@scriptcs/core it should be noted that it is a breaking API change, i.e. the addition of `IConsole.Width` so we'll need an opinion on whether we want to take this.

filipw:
take it it's a small breaking change

adamralph:
Yep, it will only break third party implementations of `IConsole`.

khellang:
If we only had https://github.com/dotnet/roslyn/issues/258 :wink:

adamralph:
This PR is waiting on two things:

1. @BlackFrog1 to revert the changes to LibLog.cs
1. @scriptcs/core to decide if we are OK with the breaking change to `IConsole`

khellang:
Personally I don't really care about breaking stuff. I don't think many people will get "broken" by adding a member to an interface.

glennblock:
Adding a member to an interface breaks all implementers. In the case of
IConsole this is likely custom hosters which may not be so bad. That is not
always the case though. Adding a member to IScriptPack for example will
have much more serious consequences.

BlackFrog1:
What is the best course of action in reverting the change to LibLog.cs? Do I take all the changes back?  Or do I just resubmit the change for just this file?

adamralph:
@BlackFrog1 yes, that looks fine now, thanks.

BlackFrog1:
I think I finally got the hang of creating PR.  Once I create cleaner PR for the items in #1032, I will delete/close this.  Give me a few days.

adamralph:
Thanks @BlackFrog1, we'll await your new PR.

PR #1039, opened 2015-05-07
made the references add/remove methods virtual #1036

### LINKED ISSUE #1036: Make ScriptExecutor references methods virtual
When hosting scriptcs and using the ScriptExecutor it'd be nice to have a way to be notified if any assemblies or namespaces change.

I see two ways of doing this.  
1) Extending the IScriptExecutor interface to have an event or action that is called.
2) Making all the ```AddReferences```, ```RemoveReferences```, etc methods virtual, so they can be overwritten with functionality that notifies.

lukebuehler:
For now I made only changes to add/remove references and left the namespaces alone. This resolves #1036

### DIFF
--- a/src/ScriptCs.Core/ScriptExecutor.cs
+++ b/src/ScriptCs.Core/ScriptExecutor.cs
@@ -94,34 +94,6 @@ namespace ScriptCs
             }
         }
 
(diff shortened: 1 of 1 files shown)


adamralph:
:+1: thanks!

PR #1043, opened 2015-05-11
 Remove IReplCommand argument getting evaluated  

BlackFrog1:
By removing the evaluation of arguments, it allows a cleaner command-line arguments.

cleaner command arguments


BlackFrog1:
I think this is my fault.  I forgot to run the test units.

glennblock:
Hi @BlackFrog1 this a functional change and should be discussed first. The
evaluation of args was by design.

PR #1044, opened 2015-05-14
Keep quoted strings together when splitting REPL Command line arguments

BlackFrog1:
Because the Repl Command line was getting split with white space, it wouldn't handle quoted text with white space.  Now you can use the "Change Directory" command with long path name.  In the screen shot below, I demonstrate missing quotes and long path names with white space.

foo 01

### DIFF
--- a/src/ScriptCs.Core/Repl.cs
+++ b/src/ScriptCs.Core/Repl.cs
@@ -55,7 +55,7 @@ namespace ScriptCs
             {
                 if (script.StartsWith(":"))
                 {
-                    var tokens = script.Split(' ');
+                    var tokens = script.SplitQuoted();
                     if (tokens[0].Length > 1)
                     {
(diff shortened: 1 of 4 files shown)


adamralph:
Thanks @BlackFrog1! IIRC this is something we chose to defer in the initial implementation but it seems to have been forgotten about since. Until now! :wink:

One thing to consider is that this is only a partial solution, e.g.

`:cd foo + bar`

will still not work even though the expression is valid C#.

This begs the question of whether the argument delimiters should also be subject to the constraints of the current language (engine). E.g. in C# this would be a comma. To do this, we'd have to be able to ask the engine to take a string (the entire contents of the command after the first space), and return a list of arguments.

glennblock:
@BlackFrog1 If you'd like to update it to the dev branch, please reopen and we'll get it in.

PR #1046, opened 2015-05-14
updated Mono.CSharp and NRefactory

filipw:
Mono.CSharp -> 4.0.0
NRefactory -> 5.5.1


khellang:
Does this mean we now have support for C# 6 with the Mono engine? :grin:

filipw:
yes indeed

screenshot 2015-05-15 11 09 12

khellang:
Nice! :sunglasses: :+1:

PR #1047, opened 2015-05-15
Word wrap the Repl Command description in the Help Command

BlackFrog1:
I have added screen-shot of the word wrapping output with different size columns.

word wrap 80
word wrap 50
word wrap 30

### DIFF
--- a/src/ScriptCs.Contracts/IConsole.cs
+++ b/src/ScriptCs.Contracts/IConsole.cs
@@ -19,5 +19,7 @@ namespace ScriptCs.Contracts
         void ResetColor();
 
         ConsoleColor ForegroundColor { get; set; }
+
+        int Width { get; }
     }
 }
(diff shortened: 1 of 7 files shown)


src/ScriptCs.Contracts/IConsole.cs
```diff
@@ -19,5 +19,7 @@ public interface IConsole
         void ResetColor();
 
         ConsoleColor ForegroundColor { get; set; }
+
+        int Width { get; }
```
adamralph: Just dropping a comment here to annotate that this is a breaking change.

glennblock:
@scriptcs/core thoughts? We're about to break a bunch of stuff including hosters so I feel like this is a good time to merge this if we want to. Otherwise we can close it if we don't think it is worth the break on IConsole.

glennblock:
OK, it's in! Thanks @BlackFrog1!

Issue #1061, opened 2015-07-02
Publish scriptcs.exe as its own nuget package


glennblock:
Per chatting with @stankovski, it would be useful to publish scriptcs.exe as its own NuGet package, thus allowing automation scripts to easily install and run the exe from a local bin folder.

---

Depends on https://github.com/scriptcs/scriptcs/pull/1064

adamralph:
:+1: the exe should sit inside /tools

glennblock:
Will the MSBuild target automatically do that if it sees an exe? I am working on this PR now, and was modifying our Build.proj and adding a new <NuGetProjects> entry.

glennblock:
I'll shortly find out I guess..I am testing :-)

glennblock:
@adamralph yep, it puts it in tools, perfect!

stankovski:
Great! Thanks a lot guys.

adamralph:
@glennblock it's going into tools because that's what nuspec is telling it to do, see https://github.com/scriptcs/scriptcs/blob/a4bbfb38dbb60db1709db1bd932f27fd2d2bd2bc/src/ScriptCs/Properties/scriptcs.nuspec

It should be noted that this is the nuspec which is designed to create the Chocolatey package. This has a couple of consequences:

* It takes a dependency on NuGet.CommandLine
* It includes a chocolateyInstall.ps1

The presence of chocolateyInstall.ps1, whilst a little ugly, isn't really much of an issue. It would be nice to remove it though.

IMO the NuGet.CommandLine dependency needs to be removed from the NuGet package. It doesn't actually break anything, but it's completely inappropriate and it just feels too odd to ship the package with this dependency. Moreover, I'm wondering if the NuGet.CommandLine dependency is even necessary for the Chocolatey package.

Which brings me to another point. Our Chocolatey package has been shipping with the Roslyn.Compilers.* assemblies baked in for some time now. AFAIK this is a (accidental?) license violation and the original design was that we download the Roslyn.Compilers.* separately during installation using nuget.exe, hence the dependency on NuGet.CommandLine. I'm not sure if this still happens or not. Perhaps it does, and overwrites the Roslyn.Compilers.* which we are shipping in our package.

[...]

PR #1062, opened 2015-07-02
Adding scriptcs NuGet package #1061

glennblock:
Fix for #1061


adamralph:
See https://github.com/scriptcs/scriptcs/issues/1061#issuecomment-118159535

glennblock:
Tested this out and it will work fine for using the nuget.exe command line, which is the main use case for its existence

PR #1067, opened 2015-07-28
Fix 'else if' bug for mono #1060

### LINKED ISSUE #1060: else keyword not correctly parsed
I was writing a script getting the following error:
```bash
(1,1): error CS1525: Unexpected symbol `else' 
```

To verify that the problem is really about the `else` keyword I've reduced it to the following snippet:
```csharp
using System;

var x = true;
if (x) {
	Console.WriteLine("IF {0}", x);
} else {
	Console.WriteLine("ELSE {0}", x);
}
```

I've installed **scriptcs** via `brew` on OS X Yosemite. I can't tell you the version using `sciptcs -Version`, since it fails printing the version.

```bash
➜  temp  brew info scriptcs
scriptcs: stable 0.14.1 (bottled)
Tools to write and execute C#
https://github.com/scriptcs/scriptcs
/usr/local/Cellar/scriptcs/0.14.1 (31 files, 9,9M) *
  Poured from bottle
From: https://github.com/Homebrew/homebrew/blob/master/Library/Formula/scriptcs.rb
==> Dependencies
Recommended: mono ✔
==> Options
--without-mono
	Build without mono support
```

If you need any other info to solve the issue, please tell me.

ztone:
This should fix #1060.

### DIFF
--- a/src/ScriptCs.Engine.Mono/Segmenter/Lexer/ScriptLexer.cs
+++ b/src/ScriptCs.Engine.Mono/Segmenter/Lexer/ScriptLexer.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.IO;
 using System.Text.RegularExpressions;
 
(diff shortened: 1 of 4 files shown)


adamralph:
@ztone legend! :clap:

glennblock:
Awesome work!

PR #1068, opened 2015-07-28
restore deprecated Common.Logging overloads #847

### LINKED ISSUE #847: Deprecate Common.Logging usage in favour of our own logging API

adamralph:
Fixes #847

This restores all the Common.Logging members as per version 0.14.1.

### DIFF
--- a/src/ScriptCs.Core/AppDomainAssemblyResolver.cs
+++ b/src/ScriptCs.Core/AppDomainAssemblyResolver.cs
@@ -14,6 +14,18 @@ namespace ScriptCs
         private readonly IAssemblyUtility _assemblyUtility;
         private readonly IDictionary<string, AssemblyInfo> _assemblyInfoMap;
 
+        [Obsolete("Support for Common.Logging types was deprecated in version 0.15.0 and will soon be removed.")]
+        public AppDomainAssemblyResolver(
+            Common.Logging.ILog logger,
+            IFileSystem fileSystem,
+            IAssemblyResolver resolver,
+            IAssemblyUtility assemblyUtility,
+            IDictionary<string, AssemblyInfo> assemblyInfoMap = null,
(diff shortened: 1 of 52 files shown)


adamralph:
Comparing the branch to 0.14.1 is a little challenging since there have been so many other changes since, but it the diff can be seen at https://github.com/scriptcs/scriptcs/compare/v0.14.1...adamralph:847

adamralph:
Incidentally, we may be able to use the ParticularLabs APIComparer to perform a static analysis of breaking changes between 0.14.1 and the 847 branch. See https://github.com/ParticularLabs/APIComparer/blob/b207b536d6898980192f695def2b503adcbf9127/APIComparer/Program.cs#L18

adamralph:
Travis is broken because of NuGet.Core renaming shenanigans. That should be fixed by https://github.com/scriptcs/scriptcs/pull/1069, which can go in before this PR.

adamralph:
This should be good to go now.

filipw:
ok great job, but i have trouble comprehending how does Autofac chooses the constructor now?

adamralph:
Dunno, it just works :stuck_out_tongue_closed_eyes: 

But actually, we don't do magic scanning and injection, right? We explicitly register types in `RuntimeServices` and wotnot? (I'm easily lost in this part of the codebase.)

khellang:
Since we don't register a `Common.Logging.ILog`, but a [`ILogProvider`](https://github.com/scriptcs/scriptcs/blob/dev/src/ScriptCs.Hosting/RuntimeServices.cs#L58), Autofac can only satisfy one ctor :smile:

filipw:
ah!

filipw:
LGTM

adamralph:
woot! thanks @filipw

PR #1071, opened 2015-07-29
restore deprecated Common.Logging.ILog MonoScriptEngine.Logger

### LINKED ISSUE #847: Deprecate Common.Logging usage in favour of our own logging API

adamralph:
fixes #847 - see https://github.com/scriptcs/scriptcs/issues/847#issuecomment-126005509


adamralph:
:astonished: that Travis build ran like lightning!

PR #1089, opened 2015-09-02
Make sure we don't use C# 6 (yet)

khellang:
Just an automated change using https://github.com/khellang/LangVersionFixer


filipw:
the fact that you, out of all people, produced an `exe` tool is very disappointing

khellang:
LOL, believe it or not, but it's still easier than requiring people to have scriptcs installed :stuck_out_tongue:

PR #1093, opened 2015-09-09
Added a hotfix which deletes temporary folders created by NuGet

### LINKED ISSUE #1053: Housekeeping of file system artifacts created by NuGet.Core
When you have a scriptcs_packages.config every time a script is run in that directory a copy of the nuget package shows up in %TEMP%/nuget and doesn't get cleaned up. We are using scriptcs to automate some server actions an found this directory was using 10GB of disk space after a couple days of running. 

Here is an example:

scriptcs_packages.config
```xml
<?xml version="1.0" encoding="utf-8"?>
<packages>
  <package id="Newtonsoft.Json" version="6.0.8" targetFramework="net45" />
</packages>
```

json.csx
```c#
using Newtonsoft.Json;

class Test {
    public int Val = 4;
}

Console.WriteLine(JsonConvert.SerializeObject(new Test()));
```

Everytime scriptcs json.csx is run a new directory shows up in %TEMP%/nuget

ChrSteinert:
This fixes #1053 temporarily until NuGet fixes this.
Because the files are not used in any manner (at least I have not found any) they can safely be removed.

### DIFF
--- a/src/ScriptCs.Hosting/Package/PackageObject.cs
+++ b/src/ScriptCs.Hosting/Package/PackageObject.cs
@@ -1,7 +1,9 @@
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using System.Runtime.Versioning;
+using System.Text.RegularExpressions;
 using NuGet;
 
 using ScriptCs.Contracts;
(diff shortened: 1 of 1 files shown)


adamralph:
Thanks!

ChrSteinert:
Ah - just saw the build fail because of a bug I introduced. Going to fix now. Please stay with me ;)

src/ScriptCs.Hosting/Package/PackageObject.cs
```diff
@@ -60,6 +62,28 @@ public IEnumerable<string> GetCompatibleDlls(FrameworkName frameworkName)
...
+            foreach (string directory in tempDirectories.Distinct())
+            {
+                if (Directory.Exists(directory))
+                {
+                    Directory.Delete(directory, true);
+                }
+            }
+
+            // /HACK
+
```
adamralph: Whilst acknowledging that this is always going to be a hack, my first question is whether this is the right place to put it.

This method retreives a list of the DLLs contained within the package which are compatible with the specified framework. Is this the right place to add the side effect of removing the folders from the temp folder? If the preceding (existing) lines in the method have the side effect of creating them, then it is the right place. If not, I think it may be worth searching for a better place to put this.

ChrSteinert: Yes, they do. Line 61 (`var dlls = _package.GetLibFiles().Where(i => i.EffectivePath.EndsWith(Dll) || i.EffectivePath.EndsWith(Exe));`) creates the temporary files. After this method, you have no access to the underlying SourcePath property. Therefore it has to be there (unfortunately).

adamralph: Wow, that is *ugly!* :astonished: 

In that case I agree, this is the right place to put this hack.

filipw: lol nuget

ChrSteinert: Indeed - this seems it. No where else the information of created temporary folders is available. Also the NuGet method does not have any other overloads for NOT creating this ...

glennblock:
Thanks @ChrSteinert! We're revving up to do a release (FINALLY)

ChrSteinert:
Awesome. Really looking forward to it!

PR #1102, opened 2015-10-17
Updated copyright to 2015

maqnouch:


khellang:
I think a copyright notice should be the year of authoring, not the current year, right? Maybe 2014-2015? :smile:

maqnouch:
Hi :bowtie:
The FSF has some helpful guidance on including a correctly dated copyright notice in your software (intended for use with the GPL, but applicable to all software):

“The copyright notice should include the year in which you finished preparing the release (so if you finished it in 1998 but didn't post it until 1999, use 1998). You should add the proper year for each release; for example, “Copyright 1998, 1999 Terry Jones” if some versions were finished in 1998 and some were finished in 1999. If several people helped write the code, use all their names.

For software with several releases over multiple years, it's okay to use a range (“2008-2010”) instead of listing individual years (“2008, 2009, 2010”) if and only if every year in the range, inclusive, really is a “copyrightable” year that would be listed individually; and you make an explicit statement in your documentation about this usage."

adamralph:
How important is it? In most projects I don't state the year. Does that
matter?

maqnouch:
Hi @adamralph  , I just watch your talk:” Getting the first PR into .NET” on YouTube: great one :+1: 
I think that we can stick to the year date of first publication, that way we avoid any confusion.

adamralph:
@maqnouch thank you, I'm glad you enjoyed the talk! :smile: 

Regarding the copyright, do you think we can remove the year completely?

maqnouch:
I just Updated my PR so the copyright year remain relevant forever.

adamralph:
Thanks @maqnouch - I squashed your commits into one and merged them to the dev branch.

I've also added you to our [contributor list](https://github.com/scriptcs/scriptcs/wiki/Contributors)! :trophy:

maqnouch:
Thanks ;)

PR #1105, opened 2015-10-21
Updated README to reflect change from packages to scriptcs_packages

ryanrousseau:
Just a small readme update. It looks like the wiki has already been updated.


adamralph:
:+1: thanks @ryanrousseau !

PR #1109, opened 2015-10-25
Openide

glennblock:
Sending the PR to give visibility and get some other eyes on it. Unit tests are still missing.

This introduces a new `:openvs` REPL command which creates a debuggable solution for scripts and then launches Visual Studio. You get full debugger support, step through, watches, etc!

Essentially it automates the approach [here] (https://github.com/scriptcs/scriptcs/wiki/Debugging-a-script) which @dschenkelman came up with. When you run the command, it will grab all scripts in the folder (including sub-folders) of the script you specified and add them to the solution.

Below you can see a screenshot of debugging a script in VS 2015 after running the command.

screen shot 2015-12-26 at 3 10 51 pm

Today this only supports Visual Studio as the project structure which it creates is not supported in VS Code, OR in Xamarin Studio. XS doesn't "seem" to have as rich support for an "exe" style project. I tried hacking the equivalent in using an XBuild project, but that didn't work. For reference, [here](https://gist.github.com/glennblock/ca4685f4679348d427f1) is the .sln created for the screenshot.

A few community requests:
* @jamesmontemagno if you have any ideas on how to do the equiv in Xamarin, I'd appreciate it! 
* @shanselman any chance of getting a flow like this to work in VS Code?
* @mat-mcloughlin Omnisharp Sublime etc?

### DIFF
--- a/src/ScriptCs.Hosting/IVisualStudioSolution.cs
+++ b/src/ScriptCs.Hosting/IVisualStudioSolution.cs
@@ -1,0 +1,15 @@
+using System;
+using System.Collections.Generic;
+
+namespace ScriptCs.Contracts
+{
+    public interface IVisualStudioSolution
+    {
+        void AddHeader();
+        void AddScriptcsProject(string scriptcsPath, string workingPath, string args, bool attach, string projectGuid);
+        void AddProject(string path, string name, string guid, string[] files);
+        void AddGlobalHeader(string projectGuid);
+        void AddGlobalNestedProjects(IList<Tuple<string, string>> nestedItems);
+        void AddGlobal(string projectGuid, IList<Tuple<string, string>> nestedItems);
+    }
+}
--- a/src/ScriptCs.Hosting/IVisualStudioSolution.cs
+++ b/src/ScriptCs.Hosting/IVisualStudioSolution.cs
@@ -6,10 +6,10 @@ namespace ScriptCs.Contracts
     public interface IVisualStudioSolution
     {
         void AddHeader();
-        void AddScriptcsProject(string scriptcsPath, string workingPath, string args, bool attach, string projectGuid);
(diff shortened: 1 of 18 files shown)


mat-mcloughlin:
Going to tag in @nosami and @david-driscoll as they can probably offer seem contributions here

glennblock:
@scriptcs/core this has been completely refactored / cleaned up and now has full unit tests coverage.

One breaking change, it introduces a `TempPath` property on `IFileSystem` which is used for grabbing the temp folder for writing the solution. This breaking change "could" be removed by moving the method to the IVisualStudioSolutionWriter, but it seemed like a gap that should be there so I added it.

glennblock:
@scriptcs/core I ended up moving everything in hosting, rather than core. The main reason is because it seemed inappropriate / heavy to put it in core.

src/ScriptCs.Hosting/IVisualStudioSolution.cs
```diff
@@ -0,0 +1,15 @@
...
+
+namespace ScriptCs.Contracts
+{
+    public interface IVisualStudioSolution
+    {
+        void AddHeader();
+        void AddScriptcsProject(string scriptcsPath, string workingPath, string args, bool attach, string projectGuid);
+        void AddProject(string path, string name, string guid, string[] files);
+        void AddGlobalHeader(string projectGuid);
+        void AddGlobalNestedProjects(IList<Tuple<string, string>> nestedItems);
```
filipw: can we get a better signature on a public member here? i.e. a descriptive type instead of a `list` of `Tuple` which is very difficult to comprehend

glennblock: Ok

glennblock:
@filipw your move mate :-)

filipw:
:+1:

PR #1119, opened 2015-11-09
Repl enhancement

glennblock:
Adds @migueldeicaza's PR and some cleanup / fixing of tests.


src/ScriptCs.Hosting/FileConsole.cs
```diff
@@ -35,9 +35,9 @@ public void WriteLine(string value)
             this.AppendLine(value);
         }
 
-        public string ReadLine()
+        public string ReadLine(string prompt)
```
filipw: why is there `prompt` if it's not being used?

glennblock: It is used.

filipw: the whole method is

```
        public string ReadLine(string prompt)
        {
            var line = _innerConsole.ReadLine("");
            this.AppendLine(line);
            return line;
        }
```

khellang: It's just not used in the `FileConsole`.

filipw: ah now i see, ok!

glennblock: Correct, it is part of the general interface

filipw:
ok let's merge this, if something is still discovered to be wrong please make a corrective PR

I wanna play with this

PR #1123, opened 2015-12-02
(docs) Removing usage of Chocolatey aliases

gep13:
While the current usage of the `cinst` and `cup` aliases continue to work, Chocolatey is moving towards the ubiquitous usage of the `choco` command.  As a result, the new guidance for installing/updating packages is:

`choco install <packagename>`
`choco upgrade <packagename>`

As you can see [here](https://chocolatey.org/packages/ScriptCs).

So I thought I would update the installation/upgrade instructions contained on your readme.md.


adamralph:
Thanks @gep13!

PR #1125, opened 2015-12-02
(docs) Added missing opening parenthesis

gep13:
Happened to notice a missing parenthesis around the name of one of the core committer, so thought I would correct it while I was here :smile:


adamralph:
Thanks @gep13!

Issue #1127, opened 2015-12-13
CI build fails because paths are too long


adamralph:
I guess something must have changed on the build agents (and you've gotta love Windows).

E.g.

http://teamcity.codebetter.com/viewLog.html?tab=buildLog&buildTypeId=Scriptcs_CI&buildId=204147

> [01:14:25][ScriptCs.Tests.Acceptance.dll] ScriptCs.Tests.Acceptance.Packages.PackageContainsAFrameworkAssemblyReference() [01.01.02] When I install a package which contains a framework assembly reference (1s)
> [01:14:26]
[ScriptCs.Tests.Acceptance.Packages.PackageContainsAFrameworkAssemblyReference() [01.01.02] When I install a package which contains a framework assembly reference] ScriptCs.Tests.Acceptance.Support.ScriptCsException : scriptcs.exe exited with code 1. The output was: Installing packages...
> ERROR: Error installing package. [PathTooLongException] The specified path, file name, or both are too long. The fully qualified file name must be less than 260 characters, and the directory name must be less than 248 characters.

adamralph:
I guess we've no choice but too shorten the scenario names somehow so that the working folders for the acceptance tests have a shorter path.

glennblock:
If that works let's do it

adamralph:
I think this probably may have gone away. Perhaps the codebetter guys shortened the working dir paths.

The current test failures are genuine and were introduced with the merge of #1134.

adamralph:
Woot! All (mergeable) pull requests are :green_heart: 

![image](https://cloud.githubusercontent.com/assets/677704/13404462/f7d81094-df19-11e5-9995-5ed031cfe2d5.png)

I worked around the problem by restricting our builds to `JetBrains-dotnet-agent02` only. (Only 'JetBrains-dotnet-agent01' seems to have the too-long path problem.)

PR #1129, opened 2015-12-28
Adds new -e and --eval options which will immediately execute a script

### LINKED ISSUE #1085: Allow passing code to scriptcs at the command line
Ex.

'''
scriptcs -exec Console.WriteLine("Hello")
'''

Anything after -exec would be the code, thus not requiring escaping quotes.

Code should execute immediately in the REPL allowing REPL commands to work.

glennblock:
This fixes #1085 

Update: _`-e` was chosen to align with the existing *nix precedent. Mono's REPL uses this and node as well. Thanks to @migueldeicaza for the suggestion!_

It adds a new switch for immediately executing loose code. Also adds a new convenience overload to the `ScriptCsExe.Run()` method.

example:

```
scriptcs -e 'Console.WriteLine("""Hello from scriptcs""");'
```

_Note_: Depending on the shell you use you may have to jump through different hoops for passing spaces / quoted strings. The example is using Powershell. To do this from the cmd prompt you would use:

```
scriptcs -e 'Console.WriteLine(\"\"\"Hello from scriptcs\"\"\");'
```

### DIFF
--- a/src/ScriptCs/Command/ExecuteLooseScriptCommand.cs
+++ b/src/ScriptCs/Command/ExecuteLooseScriptCommand.cs
@@ -5,15 +5,15 @@ using ScriptCs.Contracts;
 
 namespace ScriptCs.Command
 {
-    internal class ExecuteScriptCommand : ExecuteScriptCommandBase, IScriptCommand
+    internal class ExecuteLooseScriptCommand : ExecuteScriptCommandBase, IExecuteLooseScriptCommand
     {
-        public ExecuteScriptCommand(
-            string script, string[] scriptArgs, 
-            IFileSystem fileSystem, IScriptExecutor scriptExecutor, 
-            IScriptPackResolver scriptPackResolver, 
-            ILogProvider logProvider, 
-            IAssemblyResolver assemblyResolver, 
-            IFileSystemMigrator fileSystemMigrator, 
+        public ExecuteLooseScriptCommand(
+            string script, string[] scriptArgs,
+            IFileSystem fileSystem, IScriptExecutor scriptExecutor,
+            IScriptPackResolver scriptPackResolver,
+            ILogProvider logProvider,
+            IAssemblyResolver assemblyResolver,
+            IFileSystemMigrator fileSystemMigrator,
             IScriptLibraryComposer composer) : 
                 base(script, scriptArgs, fileSystem, scriptExecutor, scriptPackResolver, logProvider, assemblyResolver, fileSystemMigrator, composer)
         {
@@ -26,20 +26,15 @@ namespace ScriptCs.Command
                 _fileSystemMigrator.Migrate();
 
                 var assemblyPaths = Enumerable.Empty<string>();
-                var workingDirectory = _fileSystem.GetWorkingDirectory(_script);
-                if (workingDirectory != null)
-                {
-                    assemblyPaths = _assemblyResolver.GetAssemblyPaths(workingDirectory);
-                }
-
+                var workingDirectory = _fileSystem.CurrentDirectory;
+                assemblyPaths = _assemblyResolver.GetAssemblyPaths(workingDirectory);
                 _composer.Compose(workingDirectory);
 
-                _scriptExecutor.Initialize(assemblyPaths, _scriptPackResolver.GetPacks(), ScriptArgs);
+                _scriptExecutor.Initialize(assemblyPaths, _scriptPackResolver.GetPacks());
 
                 // HACK: This is a (dirty) fix for #1086. This might be a temporary solution until some further refactoring can be done. 
                 _scriptExecutor.ScriptEngine.CacheDirectory = Path.Combine(workingDirectory ?? _fileSystem.CurrentDirectory, _fileSystem.DllCacheFolder);
-                var scriptResult = _scriptExecutor.Execute(_script, ScriptArgs);
-
+                var scriptResult = _scriptExecutor.ExecuteScript(_script, ScriptArgs);
                 var commandResult = Inspect(scriptResult);
                 _scriptExecutor.Terminate();
                 return commandResult;
--- a/src/ScriptCs/ExecuteScriptCommandBase.cs
+++ b/src/ScriptCs/ExecuteScriptCommandBase.cs
@@ -1,0 +1,88 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using ScriptCs.Command;
+using ScriptCs.Contracts;
+
+namespace ScriptCs
+{
+    public abstract class ExecuteScriptCommandBase
+    {
+        protected readonly string _script;
+        protected readonly IFileSystem _fileSystem;
+        protected readonly IScriptExecutor _scriptExecutor;
+        protected readonly IScriptPackResolver _scriptPackResolver;
+        protected readonly ILog _logger;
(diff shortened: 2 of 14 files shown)


src/ScriptCs/Command/ExecuteLooseScriptCommand.cs
```diff
@@ -0,0 +1,109 @@
+﻿using System;
+using System.IO;
+using System.Linq;
+using ScriptCs.Contracts;
+
+namespace ScriptCs.Command
+{
+    internal class ExecuteLooseScriptCommand : IExecuteLooseScriptCommand
```
filipw: this is almost a copy of `ExecuteScriptCommand`, the only difference appears to be `_scriptExecutor.Execute` vs `_scriptExecutor.ExecuteScript`.
It would be better to add a common `abstract class` and share most of the logic

filipw:
@glennblock this is great, very welcome option. What I do in a scriptcs extension for scriptcs, to allow script snippet evaluation is I saved it to a temp file (lol https://github.com/filipw/vscode-scriptcs-runner/blob/master/src/extension.ts#L81) and invoke that haha so it's very welcome

Just left a comment to avoid code copying/duplication

src/ScriptCs/Command/ExecuteLooseScriptCommand.cs
glennblock: Ok sounds good.

glennblock:
Great suggestion. Glad you find it useful.

filipw:
Also, according to TC this PR has 28 failing tests, so can you please double check everything passes OK

adamralph:
@filipw it's because of this https://github.com/scriptcs/scriptcs/issues/1127

glennblock:
Yeah it is very inconsistent that this path error pops up

adamralph:
It probably depends which build agent it runs on.

adamralph:
Scratch that. The test failures are genuine. I can reproduce them locally.

They were introduced with the merge of https://github.com/scriptcs/scriptcs/pull/1134

If I build the commit on the dev branch immediately before that merge, everything is fine.

filipw:
how is that possible, given that the build status reported on that PR was green

glennblock:
@adamralph looking into this now

adamralph:
image

Only the Travis build was allowed to run before merging, TC was still to kick in. Why the tests are passing in Linux and not in Windows, I'm not sure.

glennblock:
@adamralph I see the errors. I'm going to fix in the original branch and send another PR.

glennblock:
Working on this, I think I see the issue.

glennblock:
Fixed. I'll send a new PR shortly.

glennblock:
@adamralph @filipw all tests passing now based on my last DOH fix.

glennblock:
@adamralph any idea why TC seems to be slow in picking up the changes?

glennblock:
@filipw ready to merge!

adamralph:
> @adamralph any idea why TC seems to be slow in picking up the changes?

@glennblock Codebetter CI is just slow :unamused:

glennblock:
@scriptcs/core this is waiting on someone to click merge ;-)

glennblock:
@adamralph Travis is lying

adamralph:
@glennblock I kicked off the build again and it passed.

src/ScriptCs/Command/ExecuteLooseScriptCommand.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        public override CommandResult Execute()
+        {
+            try
+            {
+                _fileSystemMigrator.Migrate();
+
+                var assemblyPaths = Enumerable.Empty<string>();
+                var workingDirectory = _fileSystem.CurrentDirectory;
+                assemblyPaths = _assemblyResolver.GetAssemblyPaths(workingDirectory);
```
adamralph: The declaration and assignment of `assemblyPaths` can be combined on this line.

src/ScriptCs/ExecuteScriptCommandBase.cs
```diff
@@ -0,0 +1,88 @@
...
+    public abstract class ExecuteScriptCommandBase
+    {
+        protected readonly string _script;
+        protected readonly IFileSystem _fileSystem;
+        protected readonly IScriptExecutor _scriptExecutor;
+        protected readonly IScriptPackResolver _scriptPackResolver;
+        protected readonly ILog _logger;
+        protected readonly IAssemblyResolver _assemblyResolver;
+        protected readonly IFileSystemMigrator _fileSystemMigrator;
+        protected readonly IScriptLibraryComposer _composer;
```
adamralph: I don't think we should be exposing protected fields. It feels like a violation of the open/closed principle.

test/ScriptCs.Tests.Acceptance/LooseScriptExecution.cs
```diff
@@ -0,0 +1,96 @@
...
+        {
+            var scenario = MethodBase.GetCurrentMethod().GetFullName();
+
+            "Given a hello world script"
+                .f(() =>
+                {
+                    directory = ScenarioDirectory.Create(scenario);
+                    if (Environment.OSVersion.Platform == PlatformID.Win32NT)
+                    {
+                        script = @"Console.WriteLine(""""""Hello World!"""""");";
```
adamralph: Why does Windows need 3 double quotes?

adamralph:
@glennblock I left a couple of comments regarding the implementation.

Regarding behaviour, I'd like to investigate the result of combinations of a script name, an eval script and the REPL switch:

| File | Script | `-R` | Result                   |
|------|--------|------|--------------------------|
|      |        |      | REPL                     |
| X    |        |      | file, exit               |
|      | X      |      | script, exit             |
| X    | X      |      | script, exit<sup>1</sup> |
|      |        | X    | REPL                     |
| X    |        | X    | file, REPL               |
|      | X      | X    | script, exit<sup>2</sup> |
| X    | X      | X    | script, exit<sup>3</sup> |

1. We are saying that a script trumps a file. That's fine, but out of interest, how easy would it be to run both in the same session? If possible I guess we'd have to say which comes first, file or script?
2. I think this is problematic. If I pass a file and `-R`, the file is executed and then I get REPL. I think it's [astonishing](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) that when pass a script and `-R` that I don't get a REPL.
3. This is a combination of points 1 and 2.

src/ScriptCs/ExecuteScriptCommandBase.cs
glennblock: Fine

src/ScriptCs/Command/ExecuteLooseScriptCommand.cs
glennblock: You are being a little anal here

test/ScriptCs.Tests.Acceptance/LooseScriptExecution.cs
glennblock: Because it does. This has been thoroughly tested and it is apparently by design.

glennblock:
I see your point, but these seem like small issues to me. The manor use case here is to allow immediately executing a command that is passed at the command line. The other combinations are nice to have, sure they bring more consistency, but still nice to have and unlikely to be used.

Using -r however is there to allow pre-loading a bunch of stuff when you come into the REPL so it executes and then you can use the loaded objects. This feels far less valuable in this context.

Issue #1133, opened 2016-02-14
rename ScriptCs.CSharp to ScriptCs.Engine.Roslyn


glennblock:

filipw:
this is done

PR #1134, opened 2016-02-17
Renames ScriptCs.CSharp to ScriptCs.Engine.Roslyn

glennblock:
Fix for #1133


common/CommonVersionInfo.cs
```diff
@@ -1,9 +1,18 @@
+//------------------------------------------------------------------------------
```
filipw: this file should be reverted, the original even says:

```
Do not manually edit this file
```

src/ScriptCs.Engine.Roslyn/CSharpModule.cs
```diff
@@ -1,6 +1,6 @@
 ﻿using ScriptCs.Contracts;
 
-namespace ScriptCs.CSharp
+namespace ScriptCs.Engine.Roslyn
 {
     [Module("csharp")]
```
filipw: the modul;e name should be reverted too

src/ScriptCs.Hosting/ModuleLoader.cs
```diff
@@ -13,7 +13,7 @@ public class ModuleLoader : IModuleLoader
     {
         internal static readonly Dictionary<string, string> DefaultCSharpModules = new Dictionary<string, string>
         {
-            {"csharp", "ScriptCs.CSharp.dll"},
+            {"csharp", "ScriptCs.Engine.Roslyn.dll"},
```
filipw: same here

filipw:
not sure what to do with `scriptcs.visualbasic`?

glennblock:
Hmm I think my changes got overwritten when I pulled Dev. I had renamed the
module already....

glennblock:
@filipw we should remove VB probably. Maybe you should release it as a sep module?

filipw:
yeah let's remove it, can you please update this branch to revert the module rename

glennblock:
@filipw thanks for the feedback and for catching the issues. All issues have been fixed and I removed the VB module.

filipw:
:+1:

glennblock:
@filipw while you are on a roll, how about checking this? https://github.com/scriptcs/scriptcs/pull/1129

PR #1135, opened 2016-02-29
Left over fixes need for rename / removal of VB

glennblock:


glennblock:
@filipw this fixes the issues

PR #1140, opened 2016-03-22
ChocolateyInstall.ps1 should not restore Roslyn packages

jrusbatch:
Previous versions required that Roslyn be restored from NuGet. This is no longer necessary, since we now distribute a more recent release of Roslyn in our package.

Steps to test:
1. Run `build.cmd`.
2. Navigate to `artifacts\Release`.
3. Run `choco install scriptcs -source .`

Scriptcs v0.16.0 should install without error.

### DIFF
--- a/src/ScriptCs/Properties/chocolateyInstall.ps1
+++ b/src/ScriptCs/Properties/chocolateyInstall.ps1
@@ -1,18 +1,5 @@
 ﻿try {
     $tools = "$(Split-Path -parent $MyInvocation.MyCommand.Definition)"
-    $nuget = "$env:ChocolateyInstall\ChocolateyInstall\nuget"
-    $nugetPath = "$tools\nugets"
-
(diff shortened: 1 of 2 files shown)


glennblock:
Checking now

glennblock:
Verified, thanks @jrusbatch!

adamralph:
Are we complying with the license for https://www.nuget.org/packages/Roslyn.Compilers.CSharp by distributing it with scriptcs?

glennblock:
The new Roslyn bits are OSS . We can distribute it.

adamralph:
OK, so this is fine for the dev branch.

Our master branch has been distributing the old Roslyn bits for some time now (with the redundant restore in the chocolatey script) but that is a separate issue.

glennblock:
Really? I thought those are the newer bits.

adamralph:
We haven't yet released a version using anything newer than the old CTP with the restrictive license.

Issue #1143, opened 2016-03-29
The Mono engine is unable to handle escape sequences in char literals


rickardp:
Consider the following scriptlet

```
using System;

class Reprocase {
    public static int SplitIt(string s)
    {
        return s.Split(new [] {' ', '\t'}, StringSplitOptions.RemoveEmptyEntries).Length;
    }
}

Console.WriteLine("Hello world");`
```

When running it with the Mono scripting engine, it fails with the following exception:

```
ERROR: [ScriptCs.Command.ExecuteScriptCommand] Script execution failed. [System.ArgumentOutOfRangeException] Index was out of range. Must be non-negative and less than the size of the collection.
Parameter name: index
  at System.ThrowHelper.ThrowArgumentOutOfRangeException () <0x19505e0 + 0x0004f> in <filename unknown>:0 
  at System.Collections.Generic.List`1[T].get_Item (Int32 index) <0x17b6770 + 0x00022> in <filename unknown>:0 
  at Mono.CSharp.Location.get_Name () <0x3b25400 + 0x0003f> in <filename unknown>:0 
  at Mono.CSharp.Location.ToString () <0x3b253c8 + 0x00013> in <filename unknown>:0 
  at Mono.CSharp.ReportPrinter.Print (Mono.CSharp.AbstractMessage msg, System.IO.TextWriter output, Boolean showFullPath) <0x3b25188 + 0x000ab> in <filename unknown>:0 
  at Mono.CSharp.StreamReportPrinter.Print (Mono.CSharp.AbstractMessage msg, Boolean showFullPath) <0x3b25128 + 0x00027> in <filename unknown>:0 
  at Mono.CSharp.Report.Error (Int32 code, Location loc, System.String error) <0x3b24ef0 + 0x0008b> in <filename unknown>:0 
  at Mono.CSharp.CSharpParser.parse () <0x3aad1f8 + 0x0027f> in <filename unknown>:0 
  at Mono.CSharp.Evaluator.ParseString (ParseMode mode, System.String input, System.Boolean& partial_input) <0x3aa4c68 + 0x0030b> in <filename unknown>:0 
  at Mono.CSharp.Evaluator.Compile (System.String input, Mono.CSharp.CompiledMethod& compiled) <0x36dded8 + 0x0018f> in <filename unknown>:0 
[...]

PR #1144, opened 2016-03-29
#1143. Char literals does not take escape sequences into account

rickardp:
See #1143. The single quoted literal (char literal) does not work when there is more than one literal char in the string, which could be true if there are escape sequences involved.

I copied the code from the double quoted literal (string literal) token and this resolves the issue I was seeing. I had to change the code slightly to make the ShouldNotFailOnIdentifyingCharactersAsToken test succeed, as it seems the lexer needs to support some broken input (I am not sure I captured the real intent of the test though).

I added a unit test to ScriptLexerTests to show the case. It should fail without the fix and succeed with the fix


glennblock:
Thanks!

PR #1150, opened 2016-04-06
A bunch of fixes for Roslyn engine (0.16.1 hotfix)

filipw:
A bunch of fixes to make `ScriptCs.Engine.Roslyn` great again (doesn't work in 0.16.0):

 - fixed the `nuspec` for `ScriptCs.Engine.Roslyn` - it went MIA in the recent solution/projects reshuffling
 - referenced the "embedded" Roslyn DLLs from the above `nuspec` so that they are correctly bundled on release
 - fixed the name of the `ScriptCs.Engine.Roslyn` nuget package
 - removed `scriptcs_packages` and added it to `.gitignore`
 - removed obsolete VB dlls
 - killed `ScriptCs.Engine.Common` and moved its code into `ScriptCs.Engine.Roslyn`
 - bumped version to 0.16.1 as all these fixes should be released as 0.16.1


glennblock:
Looks good!

PR #1156, opened 2016-04-19
Pretty Printing

gregoryyoung:
Pretty printing support for ScriptCS in REPL

(some of original changes are squashed in @glennblock commit)

```code
$ scriptcs
scriptcs (ctrl-c to exit or :help for help)
> Env.AddCustomPrinter<int>(x => "testing " + x)
Adding custom printer for Int32
> 5
testing 5
>  
```
For running scripts without repl you can also use:

> Env.Print(25)
testing 25

There is also an overload for Print< T >(T obj) that allows for Print<Base>(derived)

@glennblock and I were also discussing the possibility of extracting a Printers interface (a little odd in a test setup to pass the dependecy in).

As state before this is to open up conversation.

### DIFF
--- a/src/ScriptCs.Core/ScriptEnvironment.cs
+++ b/src/ScriptCs.Core/ScriptEnvironment.cs
@@ -26,12 +26,12 @@ namespace ScriptCs
 
         public void Print(object o)
         {
-            Console.WriteLine(_printers.GetStringFor(o));
+            _console.WriteLine(_printers.GetStringFor(o));
         }
 
         public void Print<T>(T o)
         {
-            Console.WriteLine(_printers.GetStringFor<T>(o));
+            _console.WriteLine(_printers.GetStringFor<T>(o));
         }
 
     }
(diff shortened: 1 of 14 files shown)


glennblock:
Cool!

@filipw this doesn't move to console but it does introduce a Print method on Env which can be called outside the REPL.

glennblock:
Let's get a unit of acceptance test

src/ScriptCs.Core/ScriptEnvironment.cs
```diff
@@ -1,15 +1,38 @@
...
+
+        public void AddCustomPrinter<T>(Func<T, string> printer)
+        {
+            _console.WriteLine("Adding custom printer for " + typeof(T).Name);
+            _printers.AddCustomPrinter<T>(printer);
+        }
+
+        public void Print(object o)
+        {
+            Console.WriteLine(_printers.GetStringFor(o));
```
glennblock: Shouldn't this be _console?

src/ScriptCs.Core/ScriptEnvironment.cs
```diff
@@ -1,15 +1,38 @@
...
+        }
+
+        public void Print<T>(T o)
+        {
+            Console.WriteLine(_printers.GetStringFor<T>(o));
```
glennblock: Same as above?

glennblock:
Awesome work, thanks!

Make sure to add your name to our contributors page!

filipw:
👍 thanks

Issue #225, opened 2013-05-07
Access currently running (or possibly main) script


rossipedia:
It would be very useful to have access to the currently running file in an easy and intuitive way, preferrably as a `System.IO.FileInfo` instance.

Something like:

``` C#
FileInfo thisFile = ScriptCS.CurrentScriptFile;
```

or something similar.

SuperJMN:
:+1: I think this is a very useful feature. I already needed it and I'm just starting to use ScriptCS

adamralph:
@khellang are you working on this, or should we free it up for someone to grab?

khellang:
Well, I had another go at this earlier today, but the amount of stuff I had to touch/break in order for this to work (pass `ScriptEnvironment` through layers and layers of abstractions) made me just give up. If anyone can come up with a better solution, they should just go ahead :+1:

glennblock:
What about just reading Environment.Args? The script name / path should be
there no? At least an assumption that could be made outside the REPL.

glennblock:
After a long hiatus, I am back to working on this and I "think" I have something working. This will let you access the main file that was executed in scriptcs via new `ScriptPath` variable hanging off of `Env`. Screen shot below.

@rossipedia You can easily instantiate a FileInfo from this path. Does this look it would work for you? I could also make it a FileInfo if you "really" want that.

![screen shot 2017-01-08 at 7 25 48 pm](https://cloud.githubusercontent.com/assets/141124/21756668/0a583864-d5d9-11e6-8e5c-90164c47fa51.png)

glennblock:
I am thinking to also add a Files[] array that you can access as well which will show all files that scriptcs loaded

glennblock:
File will always be the main one

rossipedia:
@glennblock this is great! A full path is totally fine, really like the array idea as well.

glennblock:
👍

glennblock:
@rossipedia both will be there now. The main script as well as any child scripts that are loaded.

PR #1192, opened 2017-01-09
Adding ScriptPath to Env

glennblock:
# DO NOT MERGE YET

Fix for #225 

Creating this PR so others can review the approach.

This adds a new ScriptInfo class that is set with the path of the file that is executed. Ultimately the path is exposed off of the ambient `Env` property. The approach minimizes signature changes as it depends on ctor injection to get the ScriptInfo class passed around.

@scriptcs/core @rossipedia appreciate your thoughts.

Yes this issue is 2 1/2 years old.....

### DIFF
--- a/src/ScriptCs.Core/ScriptInfo.cs
+++ b/src/ScriptCs.Core/ScriptInfo.cs
@@ -1,0 +1,14 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using ScriptCs.Contracts;
+
+namespace ScriptCs
+{
+    public class ScriptInfo : IScriptInfo
+    {
+        public string ScriptPath { get; set; }
+    }
+}
--- a/src/ScriptCs.Core/ScriptInfo.cs
+++ b/src/ScriptCs.Core/ScriptInfo.cs
@@ -9,6 +9,12 @@ namespace ScriptCs
 {
     public class ScriptInfo : IScriptInfo
     {
+        public ScriptInfo()
+        {
+            LoadedScripts = new List<string>();
+        }
+
         public string ScriptPath { get; set; }
-    }
+        public IList<string> LoadedScripts { get; private set; }
+    } 
 }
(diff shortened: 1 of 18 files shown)


glennblock:
Added `LoadedScripts` property which lets you access any scripts that were loaded via #load

glennblock:
Here is a screenshot of both new properties being used.

screen shot 2017-01-09 at 12 05 59 am

src/ScriptCs.Core/ScriptInfo.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+    public class ScriptInfo : IScriptInfo
+    {
+        public ScriptInfo()
+        {
+            LoadedScripts = new List<string>();
+        }
+
+        public string ScriptPath { get; set; }
+        public IList<string> LoadedScripts { get; private set; }
```
filipw: I think maybe this can even be `HashSet<string>` since there is no need to show the same script twice

filipw:
maybe we could add current script to `FilePreProcessorResult` as a standalone property, then we could avoid relying on `result.LoadedScripts.Skip(1)` (first == current script) as I have a feeling this might become a problem in the future

glennblock:
Great minds think alike, I thought about that right after :-)

glennblock:
What makes you think it would be a problem? The code is recursive and the
first thing it hits is the makn file? I thought of doing it because it felt
a little hacky, but not that it would break later.

src/ScriptCs.Core/ScriptInfo.cs
glennblock: I don't think we should use Hashset as it does not guarantee ordering. I want the order to be the load order.

glennblock:
Man Travis STILL causes us problems.....this is annoying

glennblock:
@filipw I did a refactoring so skip is no longer needed / I made everything first class in the FilePreProcessor. This provides other benefits so that there is no magic logic needed upstream.

filipw:
cool - LGTM

glennblock:
@filipw I said "DON'T MERGE" did you read the top? :-)

I wanted to add unit tests. But I can add those separately now I guess.

filipw:
haha I thought that was regarding the 1st commit, and that subsequent commits negated it 😄

glennblock:
:-) Next time I'll put checkboxes. with a title like "Filip, don't commit until these are done:" :p

PR #1197, opened 2017-01-12
Adding Env.ScriptAssembly

### LINKED ISSUE #244: Expose a reference to the current script assembly
I need this for the Nancy script pack.

Currently I'm doing `Assembly.GetCallingAssembly()`. Whilst this works, it feels fragile and it's not discoverable. I think scriptcs should abstract this away for me.

glennblock:
This feature adds a new `Env.ScriptAssembly` and fixes #244. The assembly is accessible in scripts or in the REPL.

### DIFF
--- a/src/ScriptCs.Contracts/IScriptEnvironment.cs
+++ b/src/ScriptCs.Contracts/IScriptEnvironment.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Reflection;
 using System.Text;
 using System.Threading.Tasks;
 
@@ -14,5 +15,6 @@ namespace ScriptCs.Contracts
         void Print(object o);
         string ScriptPath { get; }
         string[] LoadedScripts { get; }
+        Assembly ScriptAssembly { get; }
     }
 }
--- a/src/ScriptCs.Core/ScriptEnvironment.cs
+++ b/src/ScriptCs.Core/ScriptEnvironment.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Reflection;
 using ScriptCs.Contracts;
 
 namespace ScriptCs
@@ -45,6 +46,19 @@ namespace ScriptCs
         public string[] LoadedScripts
         {
             get { return _scriptInfo.LoadedScripts.ToArray(); }
+        }
+
+        private Assembly _scriptAssembly;
+        public Assembly ScriptAssembly
+        {
+            get
+            {
+                if (_scriptAssembly == null)
+                {
+                    _scriptAssembly = Assembly.GetCallingAssembly();
+                }
+                return _scriptAssembly;
+            }
         }
     }
 }
--- a/src/ScriptCs.Core/ScriptInfo.cs
+++ b/src/ScriptCs.Core/ScriptInfo.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Reflection;
 using System.Text;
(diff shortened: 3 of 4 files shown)


glennblock:
@scriptcs/core please review. This is ready to go. I'll merge tomorrow unless I hear back.

filipw:
I would advise not to inject custom code into the script itself.
You don't need to emit a fake class to make this work, this would work too:

```csharp
Env.SetScriptAssembly(System.Reflection.MethodBase.GetCurrentMethod().DeclaringType.Assembly);
```

However, you don't need to inject anything into the script code or have a set method on the env for assembly at all, you could just add this to the `ScriptEnv`:

```csharp
public Assembly ScriptAssembly {
	get {
		return AppDomain.CurrentDomain.GetAssemblies().FirstOrDefault(x => x.FullName.StartsWith == "ℛ");
	}
}
```

Finally, I think the proper way to do it, is to find the assembly up front - at compile time, before execution of the script. Right now in typical scenarios we don't distinguish compile and run steps which we could easily do.

This has great benefits, as it allows us to discover the `Assembly` *before* script is executed - but as soon as it's compiled, which can be of great benefit in hosting scenarios, are you scan your script then for types or methods, before the script runs.

I have written that sample code already before. it's here https://github.com/filipw/dotnet-script/blob/194ce42b357c0e6dbbc7647792228652c691f248/src/Dotnet.Script/ScriptStateExtensions.cs#L12

glennblock:
@filipw that doesn't work for Mono today as we're not using Roslyn and it is not named with the Roslyn naming. Also it's more expensive to have it iterate through loaded assemblies if there are a lot of them.

Yes I could use reflection, good point. I'll change to use that code rather than the dummy class as it is cleaner.

glennblock:
As far as the injection, we already do this for script libraries, it's not
a huge deal to push in one line that is mostly harmless.

glennblock:
The one line works perfectly, thanks!

glennblock:
@filipw check the latest. I changed it now just use `Assembly.CallingAssembly` within the `ScriptAssembly` property. A lot cheaper, and no more injection. 

Going to rebase before merging.

glennblock:
Rebased, ready to go.

glennblock:
This new version is MUCH simpler......

filipw:
you are right - this is much better. But it also begs the question why do it all, if the user can just call `Assembly.GetCallingAssembly()` in the script 😄 

Anyway, let's punch this in, but ultimately I think we need a solution like this https://github.com/filipw/dotnet-script/blob/194ce42b357c0e6dbbc7647792228652c691f248/src/Dotnet.Script/ScriptStateExtensions.cs#L12 because it gives us the assembly instance *before script is executed*, rather than only after, which is very valuable.

glennblock:
@filipw let's revisit it. As I said, my biggest concern is the cost of iterating through all the assemblies to find "the one".

"But it also begs the question why do it all, if the user can just call Assembly.GetCallingAssembly() in the script"

And this is why we have code reviews :-)

The one disadvantage of this approach (relating to your question) is `Assembly.GetCallingAssembly()` will only work if it is called from the script or a script library. It won't work if somehow an external assembly invoked it.

My first approach of injecting code solved that, because property caches the result after the first invocation, and because I was invoking it before any other code, it was guaranteed to be called the first time within the script itself.

With the new approach that's not the case. If someone outside the script programmatically got a hold of the IScriptEnvironment and called it, it would not work right (currently).

Now your suggested approach also works, and as you said it happens during compilation. The one downside I see to this is the cost of iterating through assemblies in the App Domain. I haven't profiled this, but because it will grow linearly with the number of assemblies, that concerns me. Do you disagree?

Also why all the concern about code injection? After all, this is how the C# compiler itself works :-)

Issue #1213, opened 2017-02-15
[InvalidOperationException] for Markdig and MoreLinq packages (version 0.17.0)


BryanWilhite:
Upgraded to new released via `ChocolateyGUI` and ran `scriptcs -install Newtonsoft.Json` this process completed without errors but installed several packages that were not expected (not dependencies of `Newtonsoft.Json`. Then I ran `scriptcs -install Markdig` and `scriptcs -install MoreLinq` both threw `Unable to find package` errors (see screenshot attached).
![clipboard01](https://cloud.githubusercontent.com/assets/5226627/22955169/8e8a26a2-f2cf-11e6-867d-38417b9b220d.png)

glennblock:
Thanks for reporting. Looking into this now. Glad Errors are now at least more readable this release. :-)

glennblock:
It looks like these are .NET Core / .NET Standard assemblies. For some reason it is pulling down a ton of packages. I am guessing it is due to us updating to the latest NuGet.Core and that the way we were calling NuGet before now is no longer correct. This is just a guess at this point, but I can see it is pulling the kitchen sink of packages.

glennblock:
Interesting, same is happening with 0.16.1 (previous release)

glennblock:
@scriptcs/core any ideas on this one? It looks like this behavior was in our previous release as well.

glennblock:
OK I've debugged it now, it happens within the NuGet.Core APIs so there must be a way to tell it to install for .NET 4.5 and not get all the .NET Standard packages.

glennblock:
OK making progress......

glennblock:
@BryanWilhite can you test out my 1213 branch (pr #1215). I believe I fixed it.

BryanWilhite:
@glennblock this is my first time checking out a pr branch. I have tried these so far:

```
$ git clone https://github.com/scriptcs/scriptcs.git -b glennblock:1215
$ git clone https://github.com/scriptcs/scriptcs.git -b pr/1215

```

glennblock:
You can just clone my fork

```
git clone https://github.com/glennblock/scriptcs.git
git checkout 1215
```

After you check out, you should see it checking out files. If you don't then do these 2 commands.

```
git pull origin 1215
```

Then you should be able to run the build script in `scriptcs/build.cmd`

BryanWilhite:
[...]

PR #1215, opened 2017-02-16
Adding custom package manager to allow forcing framework target version

glennblock:
Fix for #1213

### DIFF
--- a/src/ScriptCs.Core/FrameworkUtils.cs
+++ b/src/ScriptCs.Core/FrameworkUtils.cs
@@ -1,0 +1,30 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace ScriptCs
+{
+    public static class FrameworkUtils
+    {
+        private static string _frameworkName;
+        public static string FrameworkName
+        {
+            get
+            {
+                if (_frameworkName == null)
+                {
+                    //Thanks to Dave Glick for this code contribution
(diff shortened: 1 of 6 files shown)


daveaglick:
One suggestion I have would be to consider pulling the `FrameworkName` from either the current assembly or a known assembly instead of baking it in. That way you won't need to remember to change it if/when running on different frameworks.

See here for an example: https://github.com/Wyamio/Wyam/blob/353f72a84cea7c2747517e3b534fa8d8aaaa5fc2/src/core/Wyam.Configuration/NuGet/PackageInstaller.cs#L42

glennblock:
Right now we only support one, but that is a good point to consider if we end up getting support for .NET Core at some point.

glennblock:
@daveaglick does it make sense for us to upgrade to the NuGet 3 client? Currently we only support .NET 4.5, though we are going to move to require 4.6 soon.

glennblock:
@daveaglick thanks for the link, I will use that approach.

daveaglick:
So are you still on the v2 NuGet APIs? If so, I would suggest moving to v3. AFAIK, the v2 APIs aren't getting any more love. There's a lot of breakage though. I'll take a look tomorrow to see what it might take to bump you up.

glennblock:
@daveaglick updated. Works beautifully, thanks!

glennblock:
Allright, based on all MY testing (and the fact that Travis and TC pass) I am going to just go and merge this. I'll ask for forgiveness later if that was a bad idea.

PR #1223, opened 2017-02-27
Updating build script to force using VS2015 version of msbuild

glennblock:


glennblock:
This change fixed Team City and Windows builds. It adds VS 2015 as a requirement, but I think that is fine.

Issue #1220, opened 2017-02-23
Feature Request: NuGet Pragma


wekempf:
Using `scriptcs -install` seems a bit of a hack, to be honest. This results in having to maintain multiple files. Cake and other Roslyn based tools allow you to reference NuGet packages within the script itself. It would be great if scriptcs worked the same way.

```C#
#r SomeLocalAssembly
#n SomeNuGetPackage
```

The `#n` directive would install the NuGet package(s) to a local folder (probably packages) and then reference it(them).

While making this suggestion, I'll make another. Instead of using a generic "scriptcs_packages.config" file, use a "*scriptname*_packages.config" file. This would allow your scripts to be in the same directory without having conflicts with the packages used (i.e. each script can use their own NuGet package versions, and restore only has to work with that scripts packages). Or, work more like the new CSPROJ and require version numbers with the `#n` and don't use any config file.

glennblock:
Thanks for the suggestions.

We chose to have dependencies externalized from code as this was a common
established pattern that .NET as well as platforms like node, python, java,
ruby etc all adhere to.

That being said, I see value in #p or #n or something that allows you to
specify a package reference from within a script. #r should work today for
local assemblies.

Great idea on allowing a script-specific packages.config. That would
definitely be useful for when there are multiple scripts in a folder with
different dependencies.

wekempf:
For large multi-script projects external dependencies may make sense. I want to use scriptcs like a replacement for LinqPad, though, and in scenarios like that working with multiple files is more of a PITA than any sort of help. So, I'm glad you're receptive to the idea. :) Makes sense to me to support both modes, though that may lead to some confusion?

glennblock:
Cool. Yeah I think we can potentially support all 3. The one complexity
here is we'd have to have precedence rules. Like what happens if you use a
#n and you have a foo_scriptcs_packages.config and
scriptcs_packages.config. One way to simplify that would be to only allow
you to have one of the 3. An advantage of the separate configs is it allows
you to only load the packages your script needs.

wekempf:
Personally, I'd go with precedence script > script_packages.config > scriptcs_packages.config. Yes, both of the first two modes mean you only load the packages you need.

glennblock:
Precedence adds complexity since the same package could be overridden
versioning wise. Not allowing multiple simplifies the implementation,

wekempf:
[...]

PR #1224, opened 2017-02-27
Paket directive

glennblock:
For #1220 

# This is mostly a hack.

I've added support for paket directives in the REPL. This calls to the shim code which creates a paket.dependencies file, and calls to paket to install the packages and to create the load script.

### DIFF
--- a/src/ScriptCs.Core/ReferenceLineProcessor.cs
+++ b/src/ScriptCs.Core/ReferenceLineProcessor.cs
@@ -38,11 +38,11 @@ namespace ScriptCs
 
             var expandedArgument = Environment.ExpandEnvironmentVariables(argument);
 
-            if (argument.ToLower().Contains("paket:"))
+            if (argument.ToLower().StartsWith(Constants.PaketPrefix))
             {
-                if (!context.References.Contains(argument))
+                if (!context.CustomReferences.Contains(argument))
                 {
-                    context.References.Add(argument);
+                    context.CustomReferences.Add(argument);
                 }
                 return true;
             }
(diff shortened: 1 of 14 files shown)


glennblock:
@smoothdeveloper, one thing that is not working correctly right now is that the paket.dependencies file and packages are always installed in a temp folder, as it the load script. It seems like the paths / FileInfo object that are passed are ignored.

glennblock:
What does work though in the REPL, is you can use a #r like `#r "paket:nuget netwonsoft.json` and it will get installed / the load script will be generated, just not in the right place.

glennblock:
Another issue is the fs code requires that paket.exe be in a .paket folder local to the script or in a parent folder. For scriptcs this is a little weird, I'd like to be able to pass in the folder where paket.exe is located. The user shouldn't HAVE to worry about it. One option is i could copy paket.exe in for them, but that slows down startup and such.

smoothdeveloper:
> that the paket.dependencies file and packages are always installed 

I think that is expected, what is happening is that we find the file if it exist, and copy it to alter it in the temp directory. @forki can probably confirm.

>  I'd like to be able to pass in the folder where paket.exe is located

yes we will add a way to pass preliminary search paths as arguments (to be scanned first), thinking we could also handle $PATH which isn't the case yet.

That being said, the way it works now is idiomatic to paket usage in the while, so it will work for paket users.

src/ScriptCs.Core/ReferenceLineProcessor.cs
```diff
@@ -32,16 +32,27 @@ protected override BehaviorAfterCode BehaviorAfterCode
...
             var argument = GetDirectiveArgument(line);
-            var assemblyPath = Environment.ExpandEnvironmentVariables(argument);
 
-            var referencePath = _fileSystem.GetFullPath(assemblyPath);
+
+            var expandedArgument = Environment.ExpandEnvironmentVariables(argument);
+
+            if (argument.ToLower().StartsWith(Constants.PaketPrefix))
+            {
+                if (!context.CustomReferences.Contains(argument))
```
smoothdeveloper, line 43: maybe consider trim on the argument, it avoids trailing space (on both ends) to mess with duplicate handling you have in place.

smoothdeveloper (commented):
Cool stuff :)

PR #1227, opened 2017-02-28
basic logic implementation with inspection of result from PaketShim

smoothdeveloper:
@glennblock I made simple code changes to handle the result of PaketShim call.

I've added few notes on things that needs to be handled.

### DIFF
--- a/src/ScriptCs.Core/Repl.cs
+++ b/src/ScriptCs.Core/Repl.cs
@@ -192,12 +192,52 @@ namespace ScriptCs
         {
             var paketRefs = references.Where(r => r.StartsWith(Constants.PaketPrefix)).Select(r => r.Substring(Constants.PaketPrefix.Length)).ToArray();
             var refs = Path.Combine(FileSystem.CurrentDirectory, "Refs.csx");
-            var fi = new FileInfo(refs);
-            dynamic result = PaketShim.ResolveLoadScript(fi, paketRefs, s =>
+            
+            // todo: is this the .csx we are currently evaluating?
+            var scriptFileBeingProcessed = new FileInfo(refs);
+            
+            // todo: fix if running on mono
+            var isMono = false;
+
+            Func<string, string> prefixWithMonoIfNeeded =
+                commandLine => isMono ? "mono " + commandLine : commandLine;
+
+            ReferenceLoading.PaketHandler.ReferenceLoadingResult result = PaketShim.ResolveLoadScript(scriptFileBeingProcessed, paketRefs, prefixWithMonoIfNeeded);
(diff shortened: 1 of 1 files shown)


src/ScriptCs.Core/Repl.cs
```diff
@@ -192,12 +192,52 @@ private void HandlePaketReferences(IEnumerable<string> references)
         {
             var paketRefs = references.Where(r => r.StartsWith(Constants.PaketPrefix)).Select(r => r.Substring(Constants.PaketPrefix.Length)).ToArray();
             var refs = Path.Combine(FileSystem.CurrentDirectory, "Refs.csx");
-            var fi = new FileInfo(refs);
-            dynamic result = PaketShim.ResolveLoadScript(fi, paketRefs, s =>
+            
+            // todo: is this the .csx we are currently evaluating?
+            var scriptFileBeingProcessed = new FileInfo(refs);
```
glennblock, line 197: no, refs is not that file. In the REPL there may not even be a file. It depends. If you load scriptcs with -r you can specify a file for it to load the REPL with, but this is more of an edge case. I'd say in the REPL we just use the current directory as the hash and we can use a fixed filename like "repl.csx"

glennblock: Outside the REPL we do have a script name, and we can use that for the hash, though that code will appear elsewhere.

src/ScriptCs.Core/Repl.cs
```diff
@@ -192,12 +192,52 @@ private void HandlePaketReferences(IEnumerable<string> references)
...
+            
+            // todo: fix if running on mono
+            var isMono = false;
```
glennblock, line 200: We already have a function for determining mono that we can call here.

src/ScriptCs.Core/Repl.cs
```diff
@@ -192,12 +192,52 @@ private void HandlePaketReferences(IEnumerable<string> references)
...
+                // we could print those properties:
+                //packageManagerNotFound.implicitIncludeDir
+                //packageManagerNotFound.userProfile
+                Console.WriteLine("package manager not found");
+            }
+            else if (result.IsPackageResolutionFailed)
+            {
+                var packageResolutionFailed = (ReferenceLoading.PaketHandler.ReferenceLoadingResult.PackageResolutionFailed)result;
+    
+                Console.WriteLine(
```
glennblock, line 227: If there is an error, you Console.Error should be called as it will display in red in the REPL.

glennblock:
I'll take care of the issues you noted and fix it up a bit. Thanks!

PR #1230, opened 2017-03-01
Update Paket Resolver

smoothdeveloper:
* Allow passing framework and list of search directories to look for paket.exe
* All fresh changes / enhancement made in visualfsharp repository (better resolution for paket.exe, remove hack with replace .fsx, etc.)


glennblock:
👍

PR #1231, opened 2017-03-05
Paket refactoring

glennblock:


src/ScriptCs.Core/ScriptExecutor.cs
```diff
@@ -78,6 +81,8 @@ public class ScriptExecutor : IScriptExecutor
             Guard.AgainstNullArgument("logProvider", logProvider);
             Guard.AgainstNullArgument("composer", composer);
             Guard.AgainstNullArgument("scriptInfo", scriptInfo);
+            Guard.AgainstNullArgument("paketLoader", paketLoader);
```
smoothdeveloper, line 84: Not sure if using C# 6, but can consider nameof operator?

glennblock: Currently we are .NET 4.5 and not C# 6, we use this sort of guard code all over the codebase

PR #1245, opened 2017-07-24
Updating to Microsoft.Web.Xdt version 2.1.1 for issue #1244

### LINKED ISSUE #1244: ScriptCS includes "broken" version of Microsoft.Web.Xdt
I recently started using some functions from the [NuGet.Core](https://www.nuget.org/packages/NuGet.Core/) library in one of our scripts. I've previously noticed that `NuGet.Core` is already referenced by ScriptCS, but in order to resolve our own references I had to add it to the `scriptcs_packages.config` file anyway. `NuGet.Core` depends on [Microsoft.Web.Xdt](https://www.nuget.org/packages/Microsoft.Web.Xdt/), which is also included in ScriptCS.

Running this script on my own machine (Windows 10) works fine, but when i try to run it on one of our build agents (Windows Server 2012 R2), it throws the following error:

scriptcs_microsoftwebxdt

Both machines run the same version of ScriptCS, `0.17.1`.

I searched around the web a bit for possible solutions, and came across this post:
http://www.damirscorner.com/blog/posts/20140502-StrongNameValidationFailedSecurityException.html

`Microsoft.Web.Xdt 2.1.0` has been unlisted from nuget.org, and replaced by `2.1.1`. Comparing the two dlls, `2.1.1` seems to actually be almost exactly the same as `2.1.0` (even the same file version), but the former is signed correctly.

Checking the dll included with ScriptCS with sn.exe yields the following result:
scriptcs_snexe

...while the new (correct) version yields:
snexe_xmltransform

I tried explicitly selecting version `2.1.1` of `Microsoft.Web.Xdt` by adding it to the `scriptcs_packages.config` file, but ScriptCS seems to ignore this and attempt to load the included version anyway.

Note that this didn't actually cause us any trouble on our build agents until we started referencing `NuGet.Core` in our scripts.

andmos:
Updating to version `2.1.1` of `Microsoft.Web.Xdt`. The old 2.1.0 is unlisted from NuGet.org. 
For more info, see #1244 

Fixes #1244 

This should do the trick. Cheers!

### DIFF
--- a/src/ScriptCs.Hosting/ScriptCs.Hosting.csproj
+++ b/src/ScriptCs.Hosting/ScriptCs.Hosting.csproj
@@ -36,7 +36,7 @@
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="Microsoft.Web.XmlTransform">
-      <HintPath>..\..\packages\Microsoft.Web.Xdt.2.1.0\lib\net40\Microsoft.Web.XmlTransform.dll</HintPath>
+      <HintPath>..\..\packages\Microsoft.Web.Xdt.2.1.1\lib\net40\Microsoft.Web.XmlTransform.dll</HintPath>
     </Reference>
     <Reference Include="Newtonsoft.Json">
       <HintPath>..\..\packages\Newtonsoft.Json.6.0.3\lib\net45\Newtonsoft.Json.dll</HintPath>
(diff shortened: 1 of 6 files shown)


filipw:
Thank you - this should have been done a long time ago indeed!

PR #1247, opened 2017-07-30
Fix "using static" imports

### LINKED ISSUE #1184: using static directive fails to compile
Whenever I try to write a script with the `using static` directive (C# 6) I get the following error: 

> ERROR: Script compilation failed. [CompilationErrorException] error CS0246: The type or namespace name 'static System' could not be found (are you missing a using directive or an assembly reference?)

Here's the simple code I am trying to get to work with scriptcs:

```
using static System.Console;

WriteLine("Hello World!");
```

filipw:
Fixes #1184 

It's not the most elegant solution in the world, but given our current architecture it's OK.
The statements are still recognized as `usings` by our code (this ensures we don't break debugging), but we don't lift them out of the source code anymore. 

Long term we should probably just reference `Microsoft.CodeAnalysis.CSharp` from the core package and use Roslyn for all these parsings.
I needed to tweak some script library injection code because it went into the wrong place.

### DIFF
--- a/src/ScriptCs.Core/ScriptExecutor.cs
+++ b/src/ScriptCs.Core/ScriptExecutor.cs
@@ -208,21 +208,23 @@ namespace ScriptCs
             }
 
             var scriptLibrariesPreProcessorResult = LoadScriptLibraries(workingDirectory);
-            var safeInsertIndex = result.Code.IndexOf("#line");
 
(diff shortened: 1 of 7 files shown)


filipw:
Let's punch this in, it fixes a rather nasty bug!

PR #1251, opened 2017-07-31
Updated to latest xUnit (+ other testing libraries)

filipw:
Updated to xUnit 2.2.0 + updated all other libraries used in tests.
This also enables scriptcs to work with VS2017 Live Unit Testing feature.

A lot of changes due to namespace changes in the test libraries + binding redirects.


filipw:
This is now green. Additionally, I updated the build scripts

filipw:
green = good to go 😄

PR #1254, opened 2017-08-12
C# 7.1 support

filipw:
This PR introduces C# 7.1 features support (over the C# 7 which we already had).


filipw:
the test failure is a false positive here, this is good to go

PR #1263, opened 2017-11-10
upgrade to xunit 2.3.1 and xbehave 2.3.0-rc0001-build717

adamralph:


filipw:
Thanks. I'll need to port this into this branch https://github.com/scriptcs/scriptcs/tree/feature/sdk-projectsystem but for now, let's punch this in.

Issue #1260, opened 2017-11-04
Cannot build or test a fresh clone


adamralph:
I'm intending to send a PR to update to the latest xbehave RC, but I can't build or test a fresh clone.

If I build from the command line, the first problem I have is:

> The system cannot find the path specified.
> 
> *** BUILD FAILED ***

This is because the path to `MSBuild.exe` is hardcoded in build.cmd as `%ProgramFiles(x86)%\MSBuild\15.0\Bin\`, but in a typical VS 2017 installation the path is `%ProgramFiles(x86)%\\Microsoft Visual Studio\2017\{VS edition, e.g. Enterprise}\MSBuild\15.0\Bin`.

If I fix this, I then get

> NuGet.CommandLine.CommandLineException: Failed to load m
  sbuild Toolset ---> System.IO.FileNotFoundException: Could not load file or assembly 'Microsoft.Build, Version=14.0.0
  .0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The system cannot find the file spe
  cified.

If I open the solution in VS 2017 and run the tests, I get 22 failures in the `ScriptCs.Tests` project. I've tried using both the VS runner and TestDriven.net. The failure message is the same in each case:

> Test 'ScriptCs.Tests.CSharpScriptEngineTests+TheExecuteMethod.ShouldNotMarkSubmissionsAsIncompleteWhenRunningScript' failed:
> 	System.IO.FileLoadException : Could not load file or assembly 'System.Reflection.Metadata, Version=1.4.2.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)

filipw:
The binding redirects got inadvertently messed up in #1257, this is fixed in #1261.
The Windows build script has unfortunately been broken for a while now, it's on my TODO list. If you are not constrained by Windows, I recommend using the unix build script for now.

filipw:
This is going to be fixed in #1262

BlackFrog1:
@adamralph try PR #1269 

> === TEST EXECUTION SUMMARY ===
>     ScriptCs.Core.Tests           Total: 231, Errors: 0, Failed:  0, Skipped:  1, Time:  6.214s
[...]

PR #1267, opened 2018-01-05
update the build process

BlackFrog1:
The `build.proj` had an incorrect reference to the xUnit Runner for MSBuild which didn't exist.  Updated the global packages.config to pull down version 2.3.1 of xunit.runner.msbuild and correct file name of the DLL. (See [xUnit #1606 discussion](https://github.com/xunit/xunit/issues/1606))

Remove the hard-coded path to the MSBuild executable. The `build.cmd` needs to be run from the Developer Command Prompt.

When using VS2017, git was detecting the .vs folder, so I included an [improved git ignore](https://www.gitignore.io/api/visualstudio) file from https://www.gitignore.io/ . #1260


BlackFrog1:
#1268

BlackFrog1:
Also, I just realized that the mono build/test is failing because of incorrect nuget package.  Will submit a new version of the PR.

Issue #1268, opened 2018-01-05
Build Project has incorrect reference to xUnit Runner for MSBuild


BlackFrog1:
In the `Build.proj` it was change to use version 2.3.0 of xunit.runner.msbuild nuget package.  
https://github.com/scriptcs/scriptcs/blob/db30643aab32f77c4e6fb6544184468712dd83e0/build/Build.proj#L3

But the global nuget packages.config is only restoring version 2.2.0.
https://github.com/scriptcs/scriptcs/blob/0f888920190d1eb1a4d31db8240a8ba55aeb62ff/.nuget/packages.config#L4

Also, in version 2.3.0 and higher, the name of the dll was changed to include the framework version `xunit.runner.msbuild.net452.dll`  See xunit/xunit#1606

BlackFrog1:
#1267 was submitted to correct this issue and few other things.

filipw:
This was fixed in the recent batch of changes in the building process.

PR #1269, opened 2018-01-05
Update the build process version 2

BlackFrog1:
The `build.proj` had an incorrect reference to the xUnit Runner for MSBuild which didn't exist. Updated the global packages.config to pull down version 2.3.1 of xunit.runner.msbuild and correct file name of the DLL. (See xUnit/xUnit#1606 discussion)

Remove the hard-coded path to the MSBuild executable. The `build.cmd` needs to be run from the Developer Command Prompt.

When using VS2017, git was detecting the .vs folder, so I included an [improved git ignore](https://www.gitignore.io/api/visualstudio) file from https://www.gitignore.io/ . #1260 #1268


filipw:
Thanks for the PR. Ver much appreciate you doing this - personally I didn't want to put too much effort into reworking the build scripts after recent updates on the Roslyn side due to the fact that we will move to .NET SDK soon.

> Remove the hard-coded path to the MSBuild executable. The build.cmd needs to be run from the Developer Command Prompt.

this is not gonna work on codebetter CI. Currently your build script reports success, but doesn't actually do anything because it can't find msbuild

```
[09:39:10]Step 1/1: Build Script (Command Line)
[09:39:10][Step 1/1] Starting: Z:\buildAgent\work\84dc0ef1e85df67e\build.cmd /l:JetBrains.BuildServer.MSBuildLoggers.MSBuildLogger,C:\BuildAgent\plugins\dotnetplugin\bin\JetBrains.BuildServer.MSBuildLoggers.4.0.dll;Verbosity=Normal;NoSummary /noconlog /m:1
[09:39:10][Step 1/1] in directory: Z:\buildAgent\work\84dc0ef1e85df67e
[09:39:11][Step 1/1] 
[09:39:11][Step 1/1] MSBuild is not found in your PATH
[09:39:11][Step 1/1] Run build.cmd from Visual Studio Command Prompt or Developer Command Prompt
[09:39:11][Step 1/1] 
[09:39:11][Step 1/1] Process exited with code 0
```

The easiest way to locate msbuild reliably so that it runs everywhere, would be to reference `vswhere` from nuget and then run `vswhere -latest -requires Microsoft.Component.MSBuild -property installationPath` - this is how it was solved in configR for example https://github.com/config-r/config-r/blob/3eb9ae3908ca59a7c390264aa816e55d49fea4fe/build.csx#L40

filipw (changes requested):
https://github.com/scriptcs/scriptcs/pull/1269#issuecomment-355768301

BlackFrog1:
I don't mind putting in the time in making ScriptCs great.
Since nuget.exe is included in the project, I will have it bring down the package for vswhere and go from there.

BlackFrog1:
I corrected the location of the msbuild and the test is running now.
build log 01

filipw:
excellent - thanks a lot

filipw (approved):

Issue #1273, opened 2018-01-06
Cannot build with VS2015 / MSBuild 14


BlackFrog1:
Cannot compile with Visual Studio 2015 because of C# 7 language in  the dev branch.

![failed vs2015](https://user-images.githubusercontent.com/550820/34644883-7df761a8-f30d-11e7-9715-85cb6b962646.png)

https://github.com/scriptcs/scriptcs/blob/a8cf5777ddbdf820ed152360cbd8d80b8ada5705/src/ScriptCs.Contracts/ScriptCs.Contracts.csproj#L16

damianh:
Use VS2017/MSBuild 15 instead.

filipw:
closing per https://github.com/scriptcs/scriptcs/pull/1274#issuecomment-378733736

Issue #1272, opened 2018-01-06
Project debug output path is different for one project


BlackFrog1:
I working on #1260 and I came across this. Is there a technical reason why the Roslyn Engine project debug output path is set differently than all the others?  All the projects have the proper project references, so the output from each will get copy to the correct output location during the build process.  

> **ScriptCs\ScriptCs.csproj:**
> &lt;OutputPath&gt;bin\\debug\\&lt;/OutputPath&gt;
> &lt;OutputPath&gt;bin\\Release\\&lt;/OutputPath&gt;
> 
> **ScriptCs.Contracts\ScriptCs.Contracts.csproj:**
> &lt;OutputPath&gt;bin\\Debug\\&lt;/OutputPath&gt;
> &lt;OutputPath&gt;bin\\Release\\&lt;/OutputPath&gt;
>    
> **ScriptCs.Core\ScriptCs.Core.csproj:**
> &lt;OutputPath&gt;bin\\Debug\\&lt;/OutputPath&gt;
> &lt;OutputPath&gt;bin\\Release\\&lt;/OutputPath&gt;
>    
> **ScriptCs.Engine.Roslyn\ScriptCs.Engine.Roslyn.csproj:**
> &lt;OutputPath&gt;..\\..\\..\\scriptcs\\src\\Scriptcs\\bin\\debug\\&lt;/OutputPath&gt;
> &lt;OutputPath&gt;bin\\Release\\&lt;/OutputPath&gt;
> 
> **ScriptCs.Hosting\ScriptCs.Hosting.csproj:**
> &lt;OutputPath&gt;bin\\Debug\\&lt;/OutputPath&gt;
> &lt;OutputPath&gt;bin\\Release\\&lt;/OutputPath&gt;

filipw:
This was fixed with #1262

Issue #1262, opened 2017-11-05
Update to .NET SDK project format


filipw:
We should move off from the old MsBuild 14 project format to the new .NET SDK project format.
I started the work on it - there is a tracking branch here https://github.com/scriptcs/scriptcs/tree/feature/sdk-projectsystem

filipw:
OK it works and builds now on Windows and *nix. 
Next step is to write the new set of build scripts!

filipw:
Fixed with #1275

Issue #1250, opened 2017-07-31
Update to C# 7


filipw:
At the moment the entire solution is pinned to C# 5.
We could benefit from newer syntax, especially around string interpolation, expression bodied members, some tuples and nicer "out" syntax.

filipw:
Done

PR #1274, opened 2018-01-06
Fix compiling with VS2015

BlackFrog1:
#1273 #1272 

Adjust the project and code to compile with C# 6.  When #1262 is ready we can move forward with C# 7 code base.


filipw:
Thanks - actually, I now realized we moved dev branch away from VS2015 by design - this was done via this issue https://github.com/scriptcs/scriptcs/issues/1250. Sorry for introducing confusion yesterday.

I would suggest to hold off on this one, because it only introduces merge conflicts against https://github.com/scriptcs/scriptcs/tree/feature/sdk-projectsystem

filipw:
thanks for the PR - however, I will close this, because following the merge of #1275, we only want to support VS 2017 anyway

PR #1275, opened 2018-01-07
Upgraded to the .NET SDK project format

filipw:
This PR moves all projects from the old csproj format to the .NET SDK project format

Prerequisites:
 - .NET SDK 2.0+ (`dotnet-cli`)
 - VS 2017 or VS Code with C# Extension 1.14+
 - Mono 5+


filipw:
this is pretty much ready, just the dreadful `build.csproj` must be adapted or removed completely. Been worrying about this for a long time, and now the moment has come 🙄

filipw:
This PR is now good to merge. I have updated the build script to use FAKE for a more managable build script than `build.csproj` 😄 
Everything appears to work as it should, tested on my PC and Mac.

@jrusbatch can you give me permissions to the TC configuration - it needs some tweaking because it reports a failure due to misconfiguration. If not, I can also set up a new one or maybe we just use appveyor (cause it's declarative like travis)

filipw:
ping @jrusbatch

filipw:
Since I don't have access to the TC configuration, I set up Appveyor build and everything is green now.
I will merge this.

I will kick out the scriptcs-ci user from the organization, I think then the build reporting from TC should stop.

PR #1280, opened 2018-04-06
Fixed build.sh

filipw:
I managed to accidentally break `build.sh` in the last PR.
The build script on Travis didn't detect it because it had its own manual steps defined.

Therefore, this PR also updates the travis config to use `build.sh`


khellang:
Nice :sparkles:

Issue #1148, opened 2016-04-03
:openvs fails if it is run without a file param


glennblock:

nrsgit:
I implemented a fix for this created a [pull request](https://github.com/scriptcs/scriptcs/pull/1282)

PR #1282, opened 2018-04-16
Fixed bug 1148 (error when running :openvs without parameters)

nel-sam:
Fixed [ bug 1148 ](https://github.com/scriptcs/scriptcs/issues/1148) (error when running :openvs without parameters) and created a new unit test for this issue.


filipw (approved):

filipw:
Thanks, looks good! 

I am still not really sure what is the use case, because the argument is needed so that startup debugging info is created correctly, and without it the `:openvs` doesn't make much sense, but nevertheless 👍

nel-sam:
I agree about the lack of real usage case, but figured at least now it won't throw an scary error.

